Issue #20, opened 2022-10-11
[Feature Request] Does AcquireTokenOptions.CorrelationId need to be a Guid?


keegan-caruso:
Does the AcquireTokenOptions.CorrelationId need to be a Guid or would a string be sufficient?

If not set, then the default value for this will be the default value of a Guid, all zeros.

Is it strictly required that the correlation ID being a GUID is another question

jmprieur:
@keegan-caruso : le'ts have a string. It allows for hierarchical correlation ids (that ASP.NET Core uses)
@jennyf19 let's replace the GUID by a string in our next release:
https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/f77360e7de563a222134147f4c1b5214927ec7bf/src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs#L66

jennyf19:
released in 1.0.0-preview.

PR #21, opened 2022-10-16
use string instead of GUID for CorrelationId

jennyf19:
#20


jmprieur (approved):
Lgtm
Thanks @jennyf19

PR #22, opened 2022-10-17
rename authenticationOptions to ApplicationAuthenticationOptions

jennyf19:


jmprieur (approved):
LGTM

@brentschmaltz, there is a name collision with AspNetCore AuthenticationOptions ....
what do think?

PR #24, opened 2022-10-18
revert back to GUID

jennyf19:
Reverting back the change of the GUID to a string.
As this is would be a breaking change for IDownstreamWebApi options


jmprieur (approved):
LGTM
Thanks @jennyf19

Issue #37, opened 2022-11-28
[Feature Request] Address API review discussion for IDownstreamRestApi


jmprieur:
**Is your feature request related to a problem? Please describe.**

Level to make the API easier to understand.

DownstreamRestApiOptions
- [x] repetition of scopes
- [x] repetition of HttpVerb
- [x] Generic for Apps
- [x] delete
- [x] update

Serialization:
- [x] delegate for serialization
- [x] delegate for de-serialization

Issue #32, opened 2022-11-18
[Feature Request] DownstreamRestApiOptions.Clone should follow AcquireTokenOptions model


brentschmaltz:
Since DownstreamRestApiOptions is not sealed, a derived class will find it hard to implement Derived.Clone().
see: https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/3de79ae285179de07b7850c6813e80384a7470f3/src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs#L12

brentschmaltz:
https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/pull/39

PR #38, opened 2022-11-30
Jmprieur/brents feedback

jmprieur:
Addresses feedback
- addressing https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/37. 
- https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/32

image

### DIFF
--- a/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Identity.Abstractions
         public string GetApiUrl()
 #pragma warning restore CA1055 // Uri return values should not be strings
         {
-            return BaseUrl?.TrimEnd('/') + $"/{RelativePath}";
+            return BaseUrl?.TrimEnd('/') + $"/{RelativePath?.TrimStart('/')}";
         }
     }
 }
--- a/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
@@ -121,7 +121,16 @@ namespace Microsoft.Identity.Abstractions
         /// Clone the options (to be able to override them).
         /// </summary>
         /// <returns>A clone of the options.</returns>
-        public virtual AuthorizationHeaderProviderOptions Clone()
+        public AuthorizationHeaderProviderOptions Clone()
+        {
+            return CloneInternal();
+        }
+
+        /// <summary>
+        /// Clone the options (to be able to override them).
+        /// </summary>
+        /// <returns>A clone of the options.</returns>
+        protected virtual AuthorizationHeaderProviderOptions CloneInternal()
         {
             return new AuthorizationHeaderProviderOptions(this);
         }
--- a/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Identity.Abstractions
 
             BaseUrl = other.BaseUrl;
             RelativePath = other.RelativePath;
-            TokenAcquirerOptions = other.TokenAcquirerOptions.Clone();
+            AcquireTokenOptions = other.AcquireTokenOptions.Clone();
             HttpMethod = other.HttpMethod;
             CustomizeHttpRequestMessage = other.CustomizeHttpRequestMessage;
             ProtocolScheme = other.ProtocolScheme;
@@ -79,7 +79,7 @@ namespace Microsoft.Identity.Abstractions
         /// <summary>
         /// Options related to token acquisition.
         /// </summary>
-        public AcquireTokenOptions TokenAcquirerOptions
+        public AcquireTokenOptions AcquireTokenOptions
         {
             get
             {
--- a/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
@@ -24,16 +24,13 @@ namespace Microsoft.Identity.Abstractions
         }
 
         /// <summary>
-        /// Copy constructor for <see cref="DownstreamRestApiOptions"/>
+        /// Copy constructor for <see cref="AuthorizationHeaderProviderOptions"/>
         /// </summary>
         /// <param name="other">Options to copy from.</param>
         /// <exception cref="ArgumentNullException">Thrown if <paramref name="other"/> is <c>null</c>.</exception>
         public AuthorizationHeaderProviderOptions(AuthorizationHeaderProviderOptions other)
         {
-            if (other == null)
-            {
-                throw new ArgumentNullException(nameof(other));
-            }
+            _ = other ?? throw new ArgumentNullException(nameof(other));
 
             BaseUrl = other.BaseUrl;
             RelativePath = other.RelativePath;
--- a/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
@@ -41,6 +41,7 @@ namespace Microsoft.Identity.Abstractions
             HttpMethod = other.HttpMethod;
             CustomizeHttpRequestMessage = other.CustomizeHttpRequestMessage;
             ProtocolScheme = other.ProtocolScheme;
+            RequestAppToken = other.RequestAppToken;
         }
 
         /// <summary>
@@ -106,6 +107,15 @@ namespace Microsoft.Identity.Abstractions
                 _protocolScheme = string.IsNullOrEmpty(value) ? throw new ArgumentNullException(_protocolScheme) : value;
             }
         }
+
+        /// <summary>
+        /// Describes if the downstream API is called on behalf of the calling service itself
+        /// (App token) or on behalf of a user processed by the service (user token).
+        /// If <c>true</c>, the token is requested on behalf of the app. Otherwise, it on-behalf of the user.
+        /// </summary>
+        /// <remarks>This is especially usefull when the call to the downstream API is described
+        /// by configuration.</remarks>
+        public bool RequestAppToken { get;set;}
 
         /// <summary>
         /// Clone the options (to be able to override them).
--- a/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
@@ -1,0 +1,96 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Options passed-in to call downstream web APIs. To call Microsoft Graph, see rather
+    /// <c>MicrosoftGraphOptions</c> in the <c>Microsoft.Identity.Web.MicrosoftGraph</c> assembly.
+    /// </summary>
+    public class AuthorizationHeaderProviderOptions
+    {   
+        /// <summary>
+        /// Default constructor.
+        /// </summary>
+        public AuthorizationHeaderProviderOptions()
+        {
+            TokenAcquirerOptions = new AcquireTokenOptions();
+        }
+
+        /// <summary>
+        /// Copy constructor
+        /// </summary>
+        /// <param name="other">Options to copy from.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="other"/> is <c>null</c>.</exception>
+        public AuthorizationHeaderProviderOptions(AuthorizationHeaderProviderOptions other)
+        {
+            if (other == null)
+            {
+                throw new ArgumentNullException(nameof(other));
+            }
+
+            BaseUrl = other.BaseUrl;
+            RelativePath = other.RelativePath;
+            TokenAcquirerOptions = other.TokenAcquirerOptions.Clone();
+            HttpMethod = other.HttpMethod;
+            CustomizeHttpRequestMessage = other.CustomizeHttpRequestMessage;
+            ProtocolScheme = other.ProtocolScheme;
+        }
+
+        /// <summary>
+        /// Base URL for the called downstream web API. For instance <c>"https://graph.microsoft.com/beta/"</c>.
+        /// </summary>
+        public string? BaseUrl { get; set; }
+
+        /// <summary>
+        /// Path relative to the <see cref="BaseUrl"/> (for instance "me").
+        /// </summary>
+        public string RelativePath { get; set; } = string.Empty;
+
+        /// <summary>
+        /// HTTP method used to call this downstream web API (by default Get).
+        /// </summary>
+        public HttpMethod HttpMethod { get; set; } = HttpMethod.Get;
+
+        /// <summary>
+        /// Provides an opportunity for the caller app to customize the HttpRequestMessage. For example,
+        /// to customize the headers. This is called after the message was formed, including
+        /// the Authorization header, and just before the message is sent.
+        /// </summary>
+        public Action<HttpRequestMessage>? CustomizeHttpRequestMessage { get; set; }
+
+        /// <summary>
+        /// Options related to the token acquisition.
+        /// </summary>
+        public AcquireTokenOptions TokenAcquirerOptions { get; set; } = new AcquireTokenOptions();
+
+        /// <summary>
+        /// Name of the protocol scheme used to create the authorization header.
+        /// By default "Bearer"
+        /// </summary>
+        public string ProtocolScheme { set; get; } = "Bearer";
+
+        /// <summary>
+        /// Clone the options (to be able to override them).
+        /// </summary>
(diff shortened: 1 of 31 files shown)


src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptions.cs
```diff
@@ -3,106 +3,37 @@
...
-        /// </summary>
-        public string RelativePath { get; set; } = string.Empty;
-
-        /// <summary>
-        /// HTTP method used to call this downstream web API (by default Get).
-        /// </summary>
-        [DefaultValue("Get")]
-        public HttpMethod HttpMethod
+        /// <inheritdoc/>
+        public override AuthorizationHeaderProviderOptions Clone()
```
brentschmaltz, line 34: This should return a DownstreamRestApiOptions

src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            return new AuthorizationHeaderProviderOptions(this);
+        }
+
+        /// <summary>
+        /// Return the downstream web API URL.
+        /// </summary>
+        /// <returns>URL of the downstream web API.</returns>
+#pragma warning disable CA1055 // Uri return values should not be strings
+        public string GetApiUrl()
```
brentschmaltz, line 134: What is the purpose of splitting the url into BaseUrl and RelativePath?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
```diff
@@ -0,0 +1,140 @@
...
+                _httpMethod = value ?? throw new ArgumentNullException(nameof(value));
+            }
+        }
+
+        /// <summary>
+        /// Provides an opportunity for the caller app to customize the HttpRequestMessage. For example,
+        /// to customize the headers. This is called after the message was formed, including
+        /// the Authorization header, and just before the message is sent.
+        /// </summary>
+        public Action<HttpRequestMessage>? CustomizeHttpRequestMessage { get; set; }
```
brentschmaltz, line 78: Suggest verb:noun for naming actions.
HttpRequestMessageCustomizer

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.HttpMethods.tt
```diff
@@ -0,0 +1,104 @@
+﻿<#@ template debug="false" hostspecific="false" language="C#" #>
```
brentschmaltz, line 1: Help me understand the purpose of using this technology?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
```diff
@@ -12,74 +12,81 @@ namespace Microsoft.Identity.Abstractions
...
         /// or Azure Signal R, where the HttpContext is not available. In other platforms, the library
         /// will find the user from the HTTP request context.</param>
         /// <param name="content">Content to send to the REST API in the case where
-        /// <see cref="DownstreamRestApiOptions.HttpMethod"/> is <code>HttpMethod.Patch</code>, 
+        /// <see cref="AuthorizationHeaderProviderOptions.HttpMethod"/> is <code>HttpMethod.Patch</code>, 
         /// <see cref="HttpMethod.Post"/>, <see cref="HttpMethod.Put"/>.</param>
         /// <param name="cancellationToken">Cancellation token.</param>
         /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
-        Task<HttpResponseMessage> CallRestApiForUserAsync(
+        public Task<HttpResponseMessage> CallRestApiAsync(
```
brentschmaltz, line 37: I am thinking the API: CallRestApiAsync(DownstreamRestApiOptions, ClaimsPrincipal, HttpContext, CancellationToken) would be very simple to do.

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
```diff
@@ -12,74 +12,81 @@ namespace Microsoft.Identity.Abstractions
...
-        /// or Azure Signal R, where the HttpContext is not available. In other platforms, the library
-        /// will find the user from the HttpContext.</param>
+        /// <param name="content">Content to send to the REST API in the case where
+        /// <see cref="AuthorizationHeaderProviderOptions.HttpMethod"/> is <code>HttpMethod.Patch</code>, 
+        /// <see cref="HttpMethod.Post"/>, <see cref="HttpMethod.Put"/>.</param>
         /// <param name="cancellationToken">Cancellation token.</param>
-        /// <returns>A strongly typed response from the web API.</returns>
-        Task<TOutput?> GetForUserAsync<TOutput>(
+        /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
+        Task<HttpResponseMessage> CallRestApiForAppAsync(
```
brentschmaltz, line 85: There is no claims principal here, how will inbound artifacts be surfaced?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
```diff
@@ -12,74 +12,81 @@ namespace Microsoft.Identity.Abstractions
...
+        /// or Azure Signal R, where the HttpContext is not available. In other platforms, the library
+        /// will find the user from the HTTP request context.</param>
         /// <param name="content">Content to send to the REST API in the case where
-        /// <see cref="DownstreamRestApiOptions.HttpMethod"/> is <code>HttpMethod.Patch</code>, 
+        /// <see cref="AuthorizationHeaderProviderOptions.HttpMethod"/> is <code>HttpMethod.Patch</code>, 
         /// <see cref="HttpMethod.Post"/>, <see cref="HttpMethod.Put"/>.</param>
         /// <param name="cancellationToken">Cancellation token.</param>
         /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
-        Task<HttpResponseMessage> CallRestApiForAppAsync(
+        Task<HttpResponseMessage> CallRestApiForUserAsync(
```
brentschmaltz, line 64: How is this different than CallRestApiAsync?
Given there is a CallRestApiAsync, CallRestApiForUserAsync and CallRestApiForAppAsync how are these different?
If different actions are expected between CallRestApiForUserAsync and CallRestApiForAppAsync how will the implementation of CallRestApiAsync know what to do?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptions.cs
```diff
@@ -115,23 +46,26 @@ public string ProtocolScheme
...
+        /// <code>JsonSerializer.Deserialize&lt;TOutput&gt;(content, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });</code>
         /// </summary>
-        /// <returns>URL of the downstream web API.</returns>
-#pragma warning disable CA1055 // Uri return values should not be strings
-        public string GetApiUrl()
-#pragma warning restore CA1055 // Uri return values should not be strings
-        {
-            return BaseUrl?.TrimEnd('/') + $"/{RelativePath}";
-        }
+        public Func<string?, object?>? Deserializer { get; set; }
```
brentschmaltz, line 69: Why does the Deserializer take a string and the Serializer take a StringContent?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptions.cs
jmprieur: then that cannot be an `override` method, this would be to be a new method?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
jmprieur: One usage our customer wanted is base URL for the controller, and the relative path is the action.

src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptionsNoHttpMethod.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Specialization of <see cref="DownstreamRestApiOptions"/> where the <see cref="AuthorizationHeaderProviderOptions.HttpMethod"/>
+    /// is not settable beyond it's construction.
+    /// </summary>
+    /// <remarks>This class is useful to provide a better developer experience on the specialized methods
+    /// of <see cref="IDownstreamRestApi"/> where the HTTP method is provided already by the name of the 
+    /// method, and should not be overriden by the options.</remarks>
+    public class DownstreamRestApiOptionsNoHttpMethod : DownstreamRestApiOptions
```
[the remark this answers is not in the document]
jmprieur: This is what you rightly proposed. The methods which start with a verb shouldn't let the possibility of overriding the Http method?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
jmprieur: Right, I was wondering about bringing it. It does exist in the implementation.

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
jmprieur: It's calling on behalf of the apps. Would there be cases where the context is needed?

src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptions.cs
jmprieur: It's what what the HttpMessage layer does ...

src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptions.cs
jmprieur: Addressed

src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
jmprieur: Done!

src/Microsoft.Identity.Abstractions/DownstreamRestApi/DownstreamRestApiOptionsNoHttpMethod.cs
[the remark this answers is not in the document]
jmprieur:
```CSharp
           // Call a service based on the configuration only.
            downstreamRestApi.CallRestApiAsync("service");

            // Calls a service based on the programmatic description only.
            downstreamRestApi.CallRestApiAsync(null,
                options =>
                {
                    options.HttpMethod = HttpMethod.Get;
                    options.BaseUrl = "https://monApi.domain.com";
                    options.RelativePath = "api/values";
                });

            // In the following call, it's not possible to set the HttpMethod in the delegate, as it would no
            // make sense: it's already provided in the name of the method
            // The following code does not build (on purpose):
            // downstreamRestApi.DeleteForAppAsync("serviceName", "todo", options => { options.HttpMethod = HttpMethod.Put });
```

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.HttpMethods.tt
jmprieur: It was actually super useful yesterday when I changed the signature of all the verbed method with one line of change in the .tt (the serviceName shoud have been a string? from the beginning)

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
jmprieur: Added

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.HttpMethods.tt
brentschmaltz: I haven't seen this in any of our other projects, now when someone has to work in this project, that person will have to understand this technology.
If this is so helpful, we should agree across the team to use it, not in a one-off manner.

tests/DownstreamRestApiTests/CredentialDescriptionTest.cs
```diff
@@ -0,0 +1,221 @@
+﻿using Microsoft.Identity.Abstractions;
+using System.Runtime.ConstrainedExecution;
+using Xunit;
+
+namespace UnitTests
+{
+    public class CredentialDescriptionTest
+    {
+        [Fact]
+        public void CredentialDescriptionProperties()
```
brentschmaltz, line 10: This needs to be theory based.
Each test case needs to validated on its own.
This model of multiple test cases leads to 'fix one, then the next'.

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.HttpMethods.tt
jmprieur: It's amazingly helpful, but you is meta => requires to think in terms of models and code generation. It's helpful when you have a do a lot of things of the same kind. it avoids copy/paste, and also ensures consistency. It help to evolve things consistency and efficiently.
Now ... it might not be for everybody.
In the end it generates code. You don't need to touch the .tt once the code is generated, until you want to evolve things. Note the usage of partial classes here.

tests/DownstreamRestApiTests/CredentialDescriptionTest.cs
jmprieur: will do
I was tempted to use t4, but I wasn't sure you'd have liked it. Theories are similiar in spirit

src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Copy constructor for <see cref="DownstreamRestApiOptions"/>
+        /// </summary>
+        /// <param name="other">Options to copy from.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="other"/> is <c>null</c>.</exception>
+        public AuthorizationHeaderProviderOptions(AuthorizationHeaderProviderOptions other)
+        {
+            if (other == null)
```
brentschmaltz, line 33: we are starting to use this pattern for checking for nulls
_ = other ?? throw ...

src/Microsoft.Identity.Abstractions/DownstreamRestApi/AuthorizationHeaderProviderOptions.cs
```diff
@@ -0,0 +1,149 @@
...
+        public AuthorizationHeaderProviderOptions()
+        {
+        }
+
+        /// <summary>
+        /// Copy constructor for <see cref="DownstreamRestApiOptions"/>
+        /// </summary>
+        /// <param name="other">Options to copy from.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="other"/> is <c>null</c>.</exception>
+        public AuthorizationHeaderProviderOptions(AuthorizationHeaderProviderOptions other)
```
brentschmaltz, line 31: Should be protected

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
brentschmaltz: Sometimes an application needs to create an outbound artifact using artifacts from the inbound request.
Api will require the implementation to put those inbound artifacts on the HttpContext or some other location.
If we were to pass a ClaimsPrincipal, we could attach those artifacts to the ClaimsPrincipal.ClaimsIdentity.BootstrapContext.

tests/DownstreamRestApiTests/AcquireTokenResultTests.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Microsoft.Identity.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace UnitTests
```
brentschmaltz, line 9: If we use different namespaces for different tests, the test explorer will be easier to navigate.

tests/DownstreamRestApiTests/DownstreamRestApiTests.cs
```diff
@@ -0,0 +1,132 @@
...
+                RelativePath = "/api/values",
+                RequestAppToken = true
+            };
+
+            Assert.Equal("https://apitocall.domain.com/api/values", downstreamRestApiOptions.GetApiUrl());
+
+            var authorizationHeaderProviderOptions = new AuthorizationHeaderProviderOptions(downstreamRestApiOptions);
+            var authorizationHeaderProviderOptionsClone = authorizationHeaderProviderOptions.Clone();
+
+            var downstreamRestApiClone = downstreamRestApiOptions.Clone();
```
brentschmaltz, line 49: With the new option 'clone' below would be of type AuthorizationHeaderProviderOptions
```cs
AuthorizationHeaderProviderOptions authorizationHeaderProviderOptions = new DownstreamRestApiOptions();
var clone = authorizationHeaderOptions.Clone();
```

brentschmaltz (approved):

brentschmaltz:
We should merge this as its really close then we can focus on finer details

PR #40, opened 2022-12-09
cleanup after testing IdWeb 2.x

jmprieur:


src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!--This should be passed from the VSTS build-->
-    <ClientSemVer Condition="'$(ClientSemVer)' == ''">1.0.5-preview</ClientSemVer>
+    <ClientSemVer Condition="'$(ClientSemVer)' == ''">1.0.0-local</ClientSemVer>
```
jmprieur, line 4: restore the usual pattern

src/Microsoft.Identity.Abstractions/DownstreamRestApi/IDownstreamRestApi.cs
```diff
@@ -29,7 +29,7 @@ public partial interface IDownstreamRestApi
         /// <param name="cancellationToken">Cancellation token.</param>
         /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
         public Task<HttpResponseMessage> CallRestApiAsync(
-            DownstreamRestApiOptions? downstreamRestApiOptions,
+            DownstreamRestApiOptions downstreamRestApiOptions,
```
jmprieur, line 32: would not make sense to be null.

PR #41, opened 2022-12-11
organized tests, cleanup of comments, marked some api's as return of null is possible

brentschmaltz:
moved and cleaned up tests under DownstreamRestApiTest to M.I.Abstractions.Tests
each test in its own namespace
cleaned up some comments
added nuget.config
moved Directory.Build.Props to root folder, test has its own Directory.Build.Props

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+  </packageSources>
+</configuration>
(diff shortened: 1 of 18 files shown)


NuGet.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
```
jmprieur, line 5: What is it for?

jmprieur (approved):
Thanks @brentschmaltz 
LGTM

NuGet.config
brentschmaltz: Standard nuget.config that Wilson uses.

brentschmaltz:
Thanks @jmprieur we are really close :-)

PR #66, opened 2023-03-10
Jmprieur/auto decrypt2

### LINKED ISSUE #65: [Feature Request] Support a credential description for auto decrypt keys
Be able to support the following configuration:

```json
            {
                "TokenDecryptionCredentials": [
                {
                    "SourceType": "AutoDecryptKeys",
                    "DecryptKeysApplicationTenant": "mytenant.onmicrosoftonline.com"
                    "DecryptKeysProtocol" : "bearer"
                }]
            }
```

This credential does not affect the container or ReferenceOrValue properties which should not be set and should return null for the soruce type "AutoDecryptKeys".

jmprieur:
Fixes #65 (new design)

cc:  @jennyf19  @brentschmaltz @RojaEnnam @TimHannMSFT  @GeoK

### DIFF
--- a/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
+++ b/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
@@ -137,6 +137,22 @@ namespace Microsoft.Identity.Abstractions
         public string? SignedAssertionFileDiskPath { get; set; }
 
         /// <summary>
+        /// Tenant in which the application is registered. This property is only used when
+        /// <see cref="SourceType"/> is <see cref="CredentialSource.AutoDecryptKeys"/>, in order
+        /// to determine the authority to use to get a token to get the decrypt the keys. The cloud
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -136,6 +136,17 @@ public class CredentialDescription
         /// </summary>
         public string? SignedAssertionFileDiskPath { get; set; }
 
+        /// <summary>
```
jennyf19, line 139: This comment is not correct, right?

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
jmprieur: Addressed. Thanks @jennyf19

PR #82, opened 2023-05-03
Add unreferenced code attributes 

westin-m:
The implementations of the methods receiving these attributes are not trim-friendly, so the warning is being bubbled up to these base members.

### DIFF
--- a/src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Identity.Abstractions
         /// </code>
         /// </example>
 #if NET6_0_OR_GREATER
-        [RequiresUnreferencedCode("Provide source generation for TInput and/or TOutput. See https://aka.ms/ms-id-web/il-trimming/generator")]
+        [RequiresUnreferencedCode("This method's implementations also use generic types and are not trim-friendly.")]
 #endif
         public Task<TOutput?> GetForUserAsync<TOutput>(
             string? serviceName,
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
```diff
@@ -44,6 +45,9 @@ public partial interface IDownstreamApi
         ///         });
         /// </code>
         /// </example>
+#if NET6_0_OR_GREATER
+        [RequiresUnreferencedCode("Provide source generation for TOutput")]
```
jmprieur, line 49:
```suggestion
        [RequiresUnreferencedCode("Provide source generation for TOutput. See https://aka.ms/ms-id-web/il-trimming/generator")]
```

jmprieur (changes requested):
Thanks @westin-m 
Please change the comments to add . See https://aka.ms/ms-id-web/il-trimming/generator in each comment where this applies.

jennyf19 (approved):
:shipit:

jennyf19:
Remember to add the issue for this work. Thanks!

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
jmprieur: I don't see it resolved?

jmprieur:
@westin-m: I'm wondering if in IdWeb. we are doing the right thing, and should not rather use different overrides of Serialize and Deserialize taking the SourceGenerationContext.

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation#complete-program-example

Happy to sync about this

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
westin-m: I changed the template and thought it had run. The changes should be there now.

jmprieur:
@westin-m I believe this PR and the issue is now no longer needed?

jmprieur (approved):
I'm not sure the warning message is relevant any longer? we won't have them provide generators? shall we?

PR #102, opened 2023-09-20
new compatibility suppression for breaking changes

westin-m:


jennyf19 (approved):

jmprieur:
What error do we get for the 5.00 when the baseline is 4.0.1?

PR #108, opened 2023-11-23
Add support for Managed Identity

JoshLozensky:
# Add support for Managed Identity

Added ManagedIdentity as a CredentialType

## Description

Added Managed Identity to the Credential Type, Description, and Source classes. Created a new class ManagedIdentityOptions with the three relevant fields ClientId, ObjectId, and HostResourceId

Fixes #1775 #2551

### DIFF
--- a/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
@@ -33,34 +33,9 @@ namespace Microsoft.Identity.Abstractions
         {
             return new ManagedIdentityOptions
             {
-                ManagedIdentityType = this.ManagedIdentityType,
-                ClientId = this.ClientId
+                ManagedIdentityType = ManagedIdentityType,
+                ClientId = ClientId
             };
-        }
-
-        /// <summary>
-        /// Ensures any <see cref="ManagedIdentityOptions"/> object is equal if all fields hold the same values.
-        /// </summary>
-        /// <param name="obj"></param>
-        /// <returns></returns>
-        public override bool Equals(object obj)
-        {
-            if (obj == null || GetType() != obj.GetType())
-            {
-                return false;
-            }
-            ManagedIdentityOptions other = (ManagedIdentityOptions)obj;
-            return ManagedIdentityType == other.ManagedIdentityType &&
-                ClientId == other.ClientId;
-        }
-
-        /// <summary>
-        /// Returns the hash code for the instance based on the values of its fields.
-        /// </summary>
-        /// <returns>A 32-bit signed integer hash code.</returns>
-        public override int GetHashCode()
-        {
-            return new { ManagedIdentityType, ClientId }.GetHashCode();
         }
     }
 }
--- a/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
@@ -6,27 +6,27 @@ using System.ComponentModel;
 namespace Microsoft.Identity.Abstractions
 {
     /// <summary>
-    /// Data object to hold the definition of a managed identity for an application to use for authentication.<br/><br/>
+    /// Data object to hold the definition of a managed identity for an application to use for authentication.
     /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for more details.
     /// </summary>
     public class ManagedIdentityOptions
     {
         /// <summary>
         /// Gets or sets whether the <see cref="Abstractions.ManagedIdentityType"/> is system-assigned or user assigned.
-        /// Defaults to <see cref="ManagedIdentityType.SystemAssigned"/> if not set.<br/><br/>
+        /// Defaults to <see cref="ManagedIdentityType.SystemAssigned"/> if not set.
         /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for details on these two types of managed identity.
         /// </summary>
         [DefaultValue(ManagedIdentityType.SystemAssigned)]
         public ManagedIdentityType ManagedIdentityType { get; set; }
 
         /// <summary>
-        /// Gets or sets the value of the client id when <see cref="ManagedIdentityType"/> is set to
-        /// <see cref="ManagedIdentityType.UserAssigned"/>.<br/>If not set, the default value is null.
+        /// Gets or sets the value of the ClientID when <see cref="ManagedIdentityType"/> is set to
+        /// <see cref="ManagedIdentityType.UserAssigned"/>. If not set, the default value is null.
         /// </summary>
         public string? ClientId { get; set; }
 
         /// <summary>
-        /// Ensures a clone of this object will not have the same reference.
+        /// Makes a new object to avoid sharing the same reference.
         /// </summary>
         /// <returns>A new instance of <see cref="ManagedIdentityOptions"/> with the same field values.</returns>
         public ManagedIdentityOptions Clone()
--- a/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.ComponentModel;
+
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Data object to hold the definition of a managed identity for an application to use for authentication.<br/><br/>
+    /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for more details.
+    /// </summary>
+    public class ManagedIdentityOptions
+    {
+        /// <summary>
+        /// Gets or sets whether the <see cref="Abstractions.ManagedIdentityType"/> is system-assigned or user assigned.
+        /// Defaults to <see cref="ManagedIdentityType.SystemAssigned"/> if not set.<br/><br/>
+        /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for details on these two types of managed identity.
+        /// </summary>
+        [DefaultValue(ManagedIdentityType.SystemAssigned)]
+        public ManagedIdentityType ManagedIdentityType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value of the client id when <see cref="ManagedIdentityType"/> is set to
+        /// <see cref="ManagedIdentityType.UserAssigned"/>.<br/>If not set, the default value is null.
+        /// </summary>
+        public string? ClientId { get; set; }
+
+        /// <summary>
+        /// Ensures a clone of this object will not have the same reference.
+        /// </summary>
+        /// <returns>A new instance of <see cref="ManagedIdentityOptions"/> with the same field values.</returns>
+        public ManagedIdentityOptions Clone()
+        {
+            return new ManagedIdentityOptions
+            {
+                ManagedIdentityType = this.ManagedIdentityType,
+                ClientId = this.ClientId
+            };
+        }
+
+        /// <summary>
+        /// Ensures any <see cref="ManagedIdentityOptions"/> object is equal if all fields hold the same values.
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public override bool Equals(object obj)
+        {
+            if (obj == null || GetType() != obj.GetType())
+            {
+                return false;
+            }
+            ManagedIdentityOptions other = (ManagedIdentityOptions)obj;
+            return ManagedIdentityType == other.ManagedIdentityType &&
+                ClientId == other.ClientId;
+        }
+
+        /// <summary>
+        /// Returns the hash code for the instance based on the values of its fields.
+        /// </summary>
+        /// <returns>A 32-bit signed integer hash code.</returns>
+        public override int GetHashCode()
+        {
+            return new { ManagedIdentityType, ClientId }.GetHashCode();
+        }
+    }
+}
--- a/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
+++ b/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
@@ -4,7 +4,7 @@
 namespace Microsoft.Identity.Abstractions
 {
     /// <summary>
-    /// Used by <see cref="ManagedIdentityOptions"/> to specify the type of managed identity to use.<br/><br/>
+    /// Used by <see cref="ManagedIdentityOptions"/> to specify the type of managed identity to use.
     /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for more details.
     /// </summary>
     public enum ManagedIdentityType
--- a/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
+++ b/src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Used by <see cref="ManagedIdentityOptions"/> to specify the type of managed identity to use.<br/><br/>
+    /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for more details.
+    /// </summary>
+    public enum ManagedIdentityType
+    {
+        /// <summary>
+        /// The default value, indicating the managed identity to use is the one configured for the Azure resource on which the
+        /// application is running.
+        /// </summary>
+        SystemAssigned = 0,
+
+        /// <summary>
+        /// Indicates the managed identity to use is a user-assigned identity which is defined in a standalone Azure resource.
+        /// </summary>
+        UserAssigned = 1,
+    }
+}
--- a/src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
@@ -41,14 +41,14 @@ namespace Microsoft.Identity.Abstractions
         }
 
         /// <summary>
-        /// Gets the name of the options describing the confidential client application (ClientId,
+        /// Gets the name of the options describing the confidential client application (ClientID,
         /// Region, Authority, client credentials). In ASP.NET Core, the authenticatiopn options name 
         /// is the same as the authentication scheme.
         /// </summary>
         public string? AuthenticationOptionsName { get; set; }
 
         /// <summary>
-        /// Sets the correlation id to be used in the request to the STS "/token" endpoint.
+        /// Sets the correlation ID to be used in the request to the STS "/token" endpoint.
         /// </summary>
         public Guid? CorrelationId { get; set; }
 
@@ -94,15 +94,12 @@ namespace Microsoft.Identity.Abstractions
 
         /// <summary>
         /// When <see cref="ManagedIdentity"/> is set, the application uses a managed identity instead of client credentials to
-        /// acquire an app token.<br/><br/>
-        /// 
+        /// acquire an app token.
         /// The type of managed identity is defined by the <see cref="ManagedIdentityOptions.ManagedIdentityType"/> field. When
         /// using a <see cref="ManagedIdentityType.SystemAssigned"/> identity, this is the only field that needs to be set and is
-        /// set by default. However, for readability it can be useful to set explicitly<br/><br/>
-        /// 
+        /// set by default. However, for readability it can be useful to set explicitly.
         /// To use a user-assigned identity, select the <see cref="ManagedIdentityType"/> that corresponds to the
-        /// <see cref="ManagedIdentityOptions.ClientId"/> you plan to use for authentication.<br/><br/>
-        /// 
+        /// <see cref="ManagedIdentityOptions.ClientId"/> you plan to use for authentication.
         /// Using either form of managed identity requires the application to be deployed on Azure and
         /// the managed identity to be configured. For more details, check the
         /// <see href="https://aka.ms/Entra/ManagedIdentityOverview"> managed identities for Azure documentation</see>.
--- a/src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Identity.Abstractions
             Claims = other.Claims;
             PopPublicKey = other.PopPublicKey;
             PopClaim = other.PopClaim;
+            ManagedIdentity = other.ManagedIdentity?.Clone();
             LongRunningWebApiSessionKey = other.LongRunningWebApiSessionKey;
             Tenant = other.Tenant;
             UserFlow = other.UserFlow;
@@ -64,7 +65,8 @@ namespace Microsoft.Identity.Abstractions
         /// A string with one or multiple claims to request. It's a json blob (encoded or not)
         /// Normally used with Conditional Access. It receives the Claims member of the UiRequiredException.
         /// It can also be used to request specific optional claims, and for 
-        /// <see href="https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/concept-conditional-access-cloud-apps">CA Auth context</see> 
+        /// <see href="https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/concept-conditional-access-cloud-apps">
+        /// CA Auth context</see>
         /// </summary>
         public string? Claims { get; set; }
 
@@ -89,6 +91,39 @@ namespace Microsoft.Identity.Abstractions
         /// In addition to the <see cref="PopPublicKey"/>, specify the PopClaim when needed in specific POP protocols. 
         /// </summary>
         public string? PopClaim { get; set; }
+
+        /// <summary>
+        /// When <see cref="ManagedIdentity"/> is set, the application uses a managed identity instead of client credentials to
+        /// acquire an app token.<br/><br/>
+        /// 
+        /// The type of managed identity is defined by the <see cref="ManagedIdentityOptions.ManagedIdentityType"/> field. When
+        /// using a <see cref="ManagedIdentityType.SystemAssigned"/> identity, this is the only field that needs to be set and is
+        /// set by default. However, for readability it can be useful to set explicitly<br/><br/>
+        /// 
+        /// To use a user-assigned identity, select the <see cref="ManagedIdentityType"/> that corresponds to the
+        /// <see cref="ManagedIdentityOptions.ClientId"/> you plan to use for authentication.<br/><br/>
+        /// 
+        /// Using either form of managed identity requires the application to be deployed on Azure and
+        /// the managed identity to be configured. For more details, check the
+        /// <see href="https://aka.ms/Entra/ManagedIdentityOverview"> managed identities for Azure documentation</see>.
+        /// </summary>
+        /// <example>
+        /// <format type="text/markdown">
+        /// <![CDATA[
+        /// The Json fragment below describes how to use a system-assigned Managed Identity for authentication in a confidential client application :
+        /// :::code language="json" source="~/../abstractions-samples/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptions.cs" id="managedidentitysystem_json":::
+        /// 
+        /// The code below describes the same, programmatically in C#.
+        /// :::code language="csharp" source="~/../abstractions-samples/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptions.cs" id="managedidentitysystem_csharp":::
+        /// 
+        /// The Json fragment below describes how to use a user-assigned Managed Identity for authentication in a confidential client application :
+        /// :::code language="json" source="~/../abstractions-samples/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptions.cs" id="managedidentityuser_json":::
+        /// 
+        /// The code below describes the same, programmatically in C#.
+        /// :::code language="csharp" source="~/../abstractions-samples/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptions.cs" id="managedidentityuser_csharp":::
+        /// ]]></format>
+        /// </example>
+        public ManagedIdentityOptions? ManagedIdentity { get; set; }
 
         /// <summary>
         /// Key used for long running web APIs that need to call downstream web
--- a/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
+++ b/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
@@ -1,0 +1,78 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Xunit;
+
+namespace Microsoft.Identity.Abstractions.Tests
+{
+    public class AquireTokenOptionsTests
+    {
+        [Fact]
+        public void ManagedIdentitySystemAssigned()
+        {
+            // App token from a system-assigned managed identity.
+            // -------------------------------------------------
+            // https://aka.ms/Entra/ManagedIdentityOverview
+            /*
+            // <managedidentity_json>
+            {
+                "AquireTokenOptions": {
+                    "ManagedIdentity": {
+                        "ManagedIdentityType": "SystemAssigned"
+                    }
+                }
+            }
+            // </managedidentitysystem_json>
+            */
+
+            // <managedidentity_csharp>
+            AcquireTokenOptions acquireTokenOptions = new AcquireTokenOptions
+            {
+                ManagedIdentity = new ManagedIdentityOptions()
+                {
+                    // default: ManagedIdentityType = ManagedIdentityType.SystemAssigned
+                }
+            };
+            // </managedidentitysystem_csharp>
+
+            Assert.Equal(ManagedIdentityType.SystemAssigned, acquireTokenOptions.ManagedIdentity.ManagedIdentityType);
+            Assert.Null(acquireTokenOptions.ManagedIdentity.ClientId);
+        }
+
+        [Fact]
+        public void ManagedIdentityUserAssigned()
+        {
+            // App token from a user-assigned managed identity.
+            // -------------------------------------------------
+            // https://aka.ms/Entra/ManagedIdentityOverview
+            /*
+            // <managedidentityuser_json>
+            {
+                "AquireTokenOptions": {
+                    "ManagedIdentity": {
+                        "ManagedIdentityType": "UserAssigned"
+                        "ClientId": "[ClientIdForTheManagedIdentityResource]"
+                    }
+                }
+            }
+            // </managedidentityuser_json>
+            */
+
+            // <managedidentityuser_csharp>
+            ManagedIdentityOptions managedIdentityDescription = new ManagedIdentityOptions
+            {
+                ManagedIdentityType = ManagedIdentityType.UserAssigned,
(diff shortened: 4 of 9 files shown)


src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,11 @@ public AcquireTokenOptions(AcquireTokenOptions other)
         /// </summary>
         public string? PopClaim { get; set; }
 
+        /// <summary>
+        /// Specifies if the client credential is a managed identity.
```
jmprieur, line 95:
```suggestion
        /// Uses a managed identity to acquire the app token instead of the client credentials.
```

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,11 @@ public AcquireTokenOptions(AcquireTokenOptions other)
...
+        /// </summary>
+        public string? ManagedIdentity { get; set; }
```
jmprieur, line 97: This should not be a `string?` but a `ManagedIdentityOptions?`

src/Microsoft.Identity.Abstractions/ApplicationOptions/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Identity.Abstractions.ApplicationOptions
```
jmprieur, line 5: Isn't the namespace of all the other classes/interfaces  Microsoft.Identity.Abstractions?

jmprieur (changes requested):
The design is to have a property of type ManagedIdentityOptions on AcquireTokenOptions, not to change CredentialSource/CredentialDescription

src/Microsoft.Identity.Abstractions/ApplicationOptions/ManagedIdentityOptions.cs
JoshLozensky: changed

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
JoshLozensky: fixed

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,31 @@ public AcquireTokenOptions(AcquireTokenOptions other)
         /// </summary>
         public string? PopClaim { get; set; }
 
+        /// <summary>
+        /// When <see cref="AcquireTokenOptions"/> is set using <see cref="ManagedIdentityOptions"/>, it specifies the ClientId,
```
jmprieur, line 95:
```suggestion
        /// When the <see cref="ManagedIdentityOptions"/> property of <see cref="AcquireTokenOptions"/> is not null, it specifies the ClientId, 
```

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,31 @@ public AcquireTokenOptions(AcquireTokenOptions other)
         /// </summary>
         public string? PopClaim { get; set; }
 
+        /// <summary>
+        /// When <see cref="AcquireTokenOptions"/> is set using <see cref="ManagedIdentityOptions"/>, it specifies the ClientId, 
+        /// ObjectId, and HostUserId of the Azure user-assigned managed identity used to aquire an app token instead of using
```
jmprieur, line 96:
```suggestion
        /// or ObjectId, orHostUserId of the Azure user-assigned managed identity used to acquire an app token instead of using
```

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,31 @@ public AcquireTokenOptions(AcquireTokenOptions other)
...
+        /// client credentials. This requires the application to be deployed on Azure, the managed identity to be configured, and
```
jmprieur, line 97:
```suggestion
        /// client credentials. This requires the application to be deployed on Azure, the managed identity to be configured.
```

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,31 @@ public AcquireTokenOptions(AcquireTokenOptions other)
...
+        /// the workload identity federation containing the managed identity is declared in the application registration.
```
jmprieur, line 98:
```suggestion
        /// 
```

This is a different scenario,

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,31 @@ public AcquireTokenOptions(AcquireTokenOptions other)
...
+        /// The Json fragment below describes how to use a Managed Identity as a client credential in a confidential client application :
+        /// :::code language="json" source="~/../abstractions-samples/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptions.cs" id="managedidentity_json":::
+        /// 
+        /// The code below describes the same, programmatically in C#.
+        /// :::code language="csharp" source="~/../abstractions-samples/test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptions.cs" id="managedidentity_csharp":::
+        /// ]]></format>
+        /// </example>
+        /// <remarks>
+        /// If you want to use the system-assigned managed identity, just define the
+        /// <see cref="AcquireTokenOptions.ManagedIdentity"/> field so that one or more of the three parameters (ClientId,
```
jmprieur, line 114:
```suggestion
        /// <see cref="AcquireTokenOptions.ManagedIdentity"/> property and keep the three parameters (ClientId,
```

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            // Client credential from Managed Identity
+            // -------------------------------------------------
+            // https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview
+            /*
+            // <managedidentity_json>
+            {
+                "AquireTokenOptions": {
+                    "ManagedIdentity": {
+                        "ClientId": "12345"
```
jmprieur, line 18: Both should be GUIDs

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+                        "ObjectId": "67890"
+                        "HostResourceId": "https://management.azure.com/"
```
jmprieur, line 20: Better ask Bogdan or Gladwin good names for these.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+                ObjectId = "67890",
+                HostResourceId = "https://management.azure.com/"
+            };
+
+            AcquireTokenOptions acquireTokenOptions = new AcquireTokenOptions
+            {
+                ManagedIdentity = managedIdentityOptions
+            };
+            // </managedidentity_csharp>
+
```
jmprieur, line 40: Could we also have a sample for system-assigned managed identity?

jmprieur (commented):
Thanks @JoshLozensky 

There are a few things to change, but this looks good already!

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -34,6 +34,7 @@ public AcquireTokenOptions(AcquireTokenOptions other)
             Claims = other.Claims;
             PopPublicKey = other.PopPublicKey;
             PopClaim = other.PopClaim;
+            ManagedIdentity = other.ManagedIdentity;
```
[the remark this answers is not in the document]
JoshLozensky: Why wouldn't it need to be cloned if all the other properties are?

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: @bgavrilMS @gladjohn what would be good names for these?

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: Absolutely

src/Microsoft.Identity.Abstractions/ApplicationOptions/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// Gets or sets the object ID of the user-assigned managed identity.
+        /// </summary>
+        public string? ObjectId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Host Resource ID of the user-assigned managed identity.
+        /// </summary>
+        public string? HostResourceId { get; set; }
+    }
+}
```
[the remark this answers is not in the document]
JoshLozensky: Yes

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            // <managedidentity_csharp>
+            ManagedIdentityDescription managedIdentityDescription = new ManagedIdentityDescription
+            {
+                Source = ManagedIdentitySource.SystemAssigned
+            };
+
+            AcquireTokenOptions acquireTokenOptions = new AcquireTokenOptions
+            {
+                ManagedIdentity = managedIdentityDescription
```
jmprieur, line 33: this is verbose.

Developers would like to write:

```csharp
 AcquireTokenOptions acquireTokenOptions = new AcquireTokenOptions
            {
                ManagedIdentity = ManagedIdentityDescription.SystemAssigned
           }
```

where `ManagedIdentityDescription.SystemAssigned` would be a static get property returning an already initialized version

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            // App token from a system-assigned managed identity.
+            // -------------------------------------------------
+            // https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview
+            /*
+            // <managedidentity_json>
+            {
+                "AquireTokenOptions": {
+                    "ManagedIdentity": {
+                        "Source": "SystemAssigned"
```
jmprieur, line 18: Personally I would be ok with:

```json
 "AquireTokenOptions": {
      "ManagedIdentity": {}
  }
```

in the case of System.Assigned managed identity. But for a beginner, what you propose is more explicit.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        [Fact]
+        public void ManagedIdentityUserAssigned()
+        {
+            // App token from a user-assigned managed identity.
+            // -------------------------------------------------
+            // https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview
+            /*
+            // <managedidentityuser_json>
+            {
+                "AquireTokenOptions": {
```
jmprieur, line 51: With the proposal above we'd have:

```json
              "AquireTokenOptions": {
                    "ManagedIdentity": {
                        "IdType": "UserAssignedManagedIdentityClientId"
                        "Id": "[InsertRelevantValueForTheManagedIdentityResource]"
                    }
                }
```

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// Data object to hold the description of a managed identity for an application to use for authentication.
+    /// </summary>
+    public class ManagedIdentityDescription
+    {
+        /// <summary>
+        /// Gets or sets the source of the managed identity as defined in <see cref="ManagedIdentitySource"/>.
+        /// Defaults to <see cref="ManagedIdentitySource.SystemAssigned"/> if not set.
+        /// </summary>
+        [DefaultValue(ManagedIdentitySource.SystemAssigned)]
```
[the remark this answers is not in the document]
bgavrilMS: The programatic model should try to match the config here https://github.com/AzureAD/microsoft-identity-web/issues/1775 

I would think @jmprieur's proposal makes most sense. 

Note than in 99% of the cases ppl either use System Assigned MSI or User Assigned MSI + client_Id.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
bgavrilMS: I tend to agree to your proposal @jmprieur . It will also be consistent with MSAL NodeJS, which is also JSON driven.

bgavrilMS (commented):
I agree with @jmprieur 's proposals. Will review again once those are taken into consideration. 

PS: naming is hard. we had endless discussions about this in MSAL .NET and MSAL JS as well.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: If you look at the implementation this is possible as I do default any instance to SystemAssigned, I was thinking the verbosity was useful in aiding comprehension/readability, many customers may not know there are two kinds of managed identity, and this will help clue them in not only to the different kinds but also which one their code is using.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: I was worried about scalability in the future if new types of managed identities are added, new apis with different methods of access or different guids. An enum should only define one property ideally and this would define two both user assigned and the guid to use.

jmprieur: there won't be new types of managed identities :-). Either it's provided by the infrastructure, or it's provided by the developer (user). This is a partition

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: If we are trying to make things harder to get wrong being more explicit for beginners is the direction we should be trying to take no?

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
jmprieur: It's a good philosophy, @JoshLozensky 
but
In IdWeb, we tend to use another which philosophy is "pay only for what you need". Or "simple things should be simple. More complex things should be possible with small incremental effort".
The whole library is like that, and FIC with MSI (or AKS) is on this principle.

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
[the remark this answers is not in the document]
JoshLozensky: I did consider it as mentioned in another comment I kept them separate as best practice in design to have enums represent only one idea. Verbosity while minorly cumbersome is more readable. That said, I'm totally fine if the consensus is to go in the other direction.

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Data object to hold the description of a managed identity for an application to use for authentication.
+    /// </summary>
+    public class ManagedIdentityDescription
+    {
+        /// <summary>
+        /// Gets or sets the type of managed identity (and if applicable the unique id) as defined in
+        /// <see cref="ManagedIdentityType"/>.Defaults to <see cref="ManagedIdentityType.SystemAssigned"/> if not set.
+        /// </summary>
+        [DefaultValue(ManagedIdentityType.SystemAssigned)]
+        public ManagedIdentityType IdType { get; set; }
```
jmprieur, line 15: Should we initialize this property to ManagedIdentityType.SystemAssigned ? since this is the default?

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +91,42 @@ public AcquireTokenOptions(AcquireTokenOptions other)
         /// </summary>
         public string? PopClaim { get; set; }
 
+        /// <summary>
+        /// When <see cref="ManagedIdentity"/> is set, the application uses a managed identity instead of client credentials to
+        /// acquire an app token.<br/><br/>
+        /// 
+        /// The type of a managed identity is defined by the <see cref="ManagedIdentityDescription.IdType"/> field, when using a
+        /// <see cref="ManagedIdentityType.SystemAssigned"/> identity, this is the only field that needs to be set.<br/><br/>
```
jmprieur, line 99: It actually does not need to be set if this is the default?

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+                    }
+                }
+            }
+            // </managedidentitysystem_json>
+            */
+
+            // <managedidentity_csharp>
+            ManagedIdentityDescription managedIdentityDescription = new ManagedIdentityDescription
+            {
+                IdType = ManagedIdentityType.SystemAssigned
```
jmprieur, line 28: Woud not even be needed if IdType is set to SystemAssigned by default

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            };
+
+            AcquireTokenOptions acquireTokenOptions = new AcquireTokenOptions
+            {
+                ManagedIdentity = managedIdentityDescription
```
jmprieur, line 33: Let's summarize:

```csharp
AcquireTokenOptions acquireTokenOptions = new AcquireTokenOptions
            {
                ManagedIdentity = new ManagedIdentityDescription() 
                { 
                  // default: IdType = ManagedIdentityType.SystemAssigned
                }
           }
```

test/Microsoft.Identity.Abstractions.Tests/ManagedIdentityDescriptionTests.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// will depend on this.
+        /// </summary>
+        [Fact]
+        public void ManagedIdentity_NoDescriptionFieldsSet()
+        {
+            // Arrange
+            ManagedIdentityDescription description = new();
+
+            // Assert
+            Assert.Equal(ManagedIdentityType.SystemAssigned, description.IdType);
```
jmprieur, line 19: Does it work? The attribute is not enough to set the value?

jmprieur (approved):
LGTM
Thanks @JoshLozensky

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
JoshLozensky: Isn't that what the default value does? I wasn't sure if I was allowed to have a constructor since you said inheritance wasn't allowed, I wasn't sure how simple the data object class needed to be.

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
JoshLozensky: It doesn't need to be this is from what I was asking around if we should model using the more readable solution even if it could be done with less code. i.e. if I look at an appsettings file and it just says managed identity I don't necessarily know there are different types of managed identities, nor would I know what type is being used without checking the documentation.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: correct see previous comment on modeling best practices for readability

test/Microsoft.Identity.Abstractions.Tests/ManagedIdentityDescriptionTests.cs
JoshLozensky: This test does work, it just makes sure that even if no values are set the ManagedIdentityDescription object will always have a type because guaranteeing that  will simplify the if we do in Id Web... I'll add some prevention tomorrow to keep anyone from setting it to null themselves.

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Text;
+
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Used by 
+    /// </summary>
+    public enum ManagedIdentityType
+
+    {
```
jennyf19, line 12: rmv line

jennyf19:
> using System;

are all the usings needed?

---
In reply to: [1835478327](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/pull/108#issuecomment-1835478327) [](http://example.com/codeflow?ancestors=1835478327)

---
Refers to: src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs:1 in 11fde20. [](commit_id = 11fde206379775f9e8e544b82cbd08d55b49a816, deletion_comment = False)

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
jennyf19: The property still needs to be set the default value is for the editor/tools.

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
```diff
@@ -0,0 +1,63 @@
...
+
+
+        /// <summary>
+        /// Gets or sets the value of the unique id correlated to the <see cref="IdType"/> field for the
+        /// managed identity resource being described. Only used when <see cref="IdType"/> is set to a user-assigned option.
+        /// </summary>
+        public string? IdValue { get; set; }
```
jennyf19, line 22: >IdValue [](http://example.com/codeflow?start=23&length=7)

What is `IdValue`? Is this a client ID? Object ID?
Seems like we have deviated from the original design from the GitHub issue, was that the intent? Seems less straightforward and more complicated.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            // App token from a user-assigned managed identity.
+            // -------------------------------------------------
+            // https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview
+            /*
+            // <managedidentityuser_json>
+            {
+                "AquireTokenOptions": {
+                    "ManagedIdentity": {
+                        "IdType": "UserAssignedClientId"
+                        "IdValue": "[InsertRelevantValueForTheManagedIdentityResource]"
```
jennyf19, line 53: sorry, I'm still not understanding what `IdType` and `IdValue` are.

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
JoshLozensky: but the test passes where I init with no values, is that not the behavior I could expect in real application?

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
JoshLozensky: If you look at the ManagedIdentityType enum the client Id is specified there for the user-assigned identities. I originally had this as two enums one for system/user-assigned and one for UniqueIdType but @jmprieur preferred a less verbose definition so we flattened them.

It deviates from the original design on purpose (I did discuss with JM) as the original design would have left too much room for misconfiguration and would also cause a messier implementation in Id Web. This will be harder to get wrong and easier to check.

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: I like this it is even more informative! :)

test/Microsoft.Identity.Abstractions.Tests/AquireTokenOptionsTests.cs
JoshLozensky: design changed no longer relevant

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityDescription.cs
JoshLozensky: design changed, no longer relevant

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            return new ManagedIdentityOptions
+            {
+                ManagedIdentityType = this.ManagedIdentityType,
+                ClientId = this.ClientId
+            };
+        }
+
+        /// <summary>
+        /// Ensures any <see cref="ManagedIdentityOptions"/> object is equal if all fields hold the same values.
```
keegan-caruso, line 42: "Ensures" probably isn't the right word here. Can we just use inheritdoc?

Also param and returns are empty

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+            ManagedIdentityOptions other = (ManagedIdentityOptions)obj;
+            return ManagedIdentityType == other.ManagedIdentityType &&
+                ClientId == other.ClientId;
+        }
+
+        /// <summary>
+        /// Returns the hash code for the instance based on the values of its fields.
+        /// </summary>
+        /// <returns>A 32-bit signed integer hash code.</returns>
```
keegan-caruso, line 60: another place we probably just want to use inheritdoc

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+        public override int GetHashCode()
+        {
+            return new { ManagedIdentityType, ClientId }.GetHashCode();
```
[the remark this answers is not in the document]
JoshLozensky: HashCode.Combine doesn't work in .NET Framework, this is the recommended way to do it as a result.

It is creating a hashcode off of a newly created object. I see your point I'll remove it. I only had it in because I was defining equals and the IDE nagged me XP

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Ensures a clone of this object will not have the same reference.
+        /// </summary>
+        /// <returns>A new instance of <see cref="ManagedIdentityOptions"/> with the same field values.</returns>
+        public ManagedIdentityOptions Clone()
+        {
+            return new ManagedIdentityOptions
+            {
+                ManagedIdentityType = ManagedIdentityType,
```
jmprieur, line 36: does the compiler understand that this is the same property?

test/Microsoft.Identity.Abstractions.Tests/DownstreamApiTests.cs
```diff
@@ -75,14 +76,16 @@ public void CloneClonesAllProperties()
             Assert.Equal(downstreamApiOptions.AcquireTokenOptions.ExtraQueryParameters, downstreamApiClone.AcquireTokenOptions.ExtraQueryParameters);
             Assert.Equal(downstreamApiOptions.AcquireTokenOptions.ForceRefresh, downstreamApiClone.AcquireTokenOptions.ForceRefresh);
             Assert.Equal(downstreamApiOptions.AcquireTokenOptions.LongRunningWebApiSessionKey, downstreamApiClone.AcquireTokenOptions.LongRunningWebApiSessionKey);
+            Assert.Equal(downstreamApiOptions.AcquireTokenOptions.ManagedIdentity.ManagedIdentityType, downstreamApiClone.AcquireTokenOptions.ManagedIdentity.ManagedIdentityType);
```
jmprieur, line 79: @JoshLozensky this needs to be fixed (possible null reference)

jmprieur (commented):
LGTM
Thanks @JoshLozensky 

A few more fixes to do (per the compiler)

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Gets or sets whether the <see cref="Abstractions.ManagedIdentityType"/> is system-assigned or user assigned.
+        /// Defaults to <see cref="ManagedIdentityType.SystemAssigned"/> if not set.<br/><br/>
+        /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for details on these two types of managed identity.
+        /// </summary>
+        [DefaultValue(ManagedIdentityType.SystemAssigned)]
+        public ManagedIdentityType ManagedIdentityType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value of the client id when <see cref="ManagedIdentityType"/> is set to
```
jennyf19, line 23: nit, please uppercase all "IDs" as in client ID, object ID, etc...

jennyf19: except in the property

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Used by <see cref="ManagedIdentityOptions"/> to specify the type of managed identity to use.<br/><br/>
+    /// See <see href="https://aka.ms/Entra/ManagedIdentityOverview"/> for more details.
+    /// </summary>
+    public enum ManagedIdentityType
+    {
+        /// <summary>
+        /// The default value, indicating the managed identity to use is the one configured for the Azure resource on which the
```
jennyf19, line 13: EDAP or Azure resource? @jmprieur ?

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +92,39 @@ public AcquireTokenOptions(AcquireTokenOptions other)
         /// </summary>
         public string? PopClaim { get; set; }
 
+        /// <summary>
+        /// When <see cref="ManagedIdentity"/> is set, the application uses a managed identity instead of client credentials to
+        /// acquire an app token.<br/><br/>
+        /// 
+        /// The type of managed identity is defined by the <see cref="ManagedIdentityOptions.ManagedIdentityType"/> field. When
+        /// using a <see cref="ManagedIdentityType.SystemAssigned"/> identity, this is the only field that needs to be set and is
+        /// set by default. However, for readability it can be useful to set explicitly<br/><br/>
```
jennyf19, line 101: remove `<br/><br/>`

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +92,39 @@ public AcquireTokenOptions(AcquireTokenOptions other)
         /// </summary>
         public string? PopClaim { get; set; }
 
+        /// <summary>
+        /// When <see cref="ManagedIdentity"/> is set, the application uses a managed identity instead of client credentials to
+        /// acquire an app token.<br/><br/>
```
jennyf19, line 97: remove `<br/><br/>`

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
```diff
@@ -90,6 +92,39 @@ public AcquireTokenOptions(AcquireTokenOptions other)
...
+        /// 
+        /// The type of managed identity is defined by the <see cref="ManagedIdentityOptions.ManagedIdentityType"/> field. When
+        /// using a <see cref="ManagedIdentityType.SystemAssigned"/> identity, this is the only field that needs to be set and is
+        /// set by default. However, for readability it can be useful to set explicitly<br/><br/>
+        /// 
+        /// To use a user-assigned identity, select the <see cref="ManagedIdentityType"/> that corresponds to the
+        /// <see cref="ManagedIdentityOptions.ClientId"/> you plan to use for authentication.<br/><br/>
```
jennyf19, line 104: remove `<br/><br/>`

test/Microsoft.Identity.Abstractions.Tests/DownstreamApiTests.cs
jennyf19, line 79: there are some warnings which need to be addressed

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityType.cs
jmprieur: yes, managed identity are configured on Azure resources (computes) in general

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
JoshLozensky: The break tags are to make the explanation easier to read in the IDE hover-over text, is there a reason we should remove them?

With breaks:
IdeTipWithBreaks

Without breaks:
IdeTipWithoutBreaks

jennyf19 (approved):
Approving as @JoshLozensky  and I discussed offline on the few remaining comments.
:shipit:

src/Microsoft.Identity.Abstractions/TokenAcquisition/AcquireTokenOptions.cs
jmprieur: I didn't know about that.
Thanks for educating us, @JoshLozensky

JoshLozensky: synced offline, for consistency and how this impacts other docs, will remove.

src/Microsoft.Identity.Abstractions/ManagedIdentity/ManagedIdentityOptions.cs
JoshLozensky: Yes, I originally had it as this.ManagedIdentityType but VS told me that was unnecessary.

jmprieur (approved):

PR #116, opened 2024-01-11
update langVersion

kellyyangsong:
Edit: discussion below yielded that we could in fact update to 12.

Could only update to 11 (not 12) since this is targeting .NET 7

image
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -5,7 +5,6 @@
   </PropertyGroup>
 
 	<PropertyGroup>
-		<TargetFrameworks>net8.0</TargetFrameworks>
 		<MicrosoftNetTestSdkVersion>17.3.2</MicrosoftNetTestSdkVersion>
 		<!-- GHSA-5crp-9r3c-p9vr -->
 		<NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
 	<PropertyGroup>
+		<TargetFrameworks>net8.0</TargetFrameworks>
 		<MicrosoftNetTestSdkVersion>17.3.2</MicrosoftNetTestSdkVersion>
 		<!-- GHSA-5crp-9r3c-p9vr -->
 		<NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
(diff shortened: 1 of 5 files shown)


keegan-caruso (changes requested):
What about this one: https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/Directory.Build.props#L28

test/Microsoft.Identity.Abstractions.Tests/Microsoft.Identity.Abstractions.Tests.csproj
```diff
@@ -4,7 +4,7 @@
     <TargetFrameworks>net7.0;net462</TargetFrameworks>
     <Nullable>enable</Nullable>
     <IsPackable>false</IsPackable>
-	<LangVersion>9.0</LangVersion>
+	<LangVersion>11</LangVersion>
```
jennyf19, line 7: we should add net8.0 instead of 7 for the tests

jennyf19:
> What about this one: https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/Directory.Build.props#L28

Yes, we should be able to do 12 here.

kellyyangsong:
> What about this one: https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/Directory.Build.props#L28

Since it's targeting [`netstandard2.0;netstandard2.1;net462`](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/aea0d8dff223f1402788e6d8dcea527c17ac028a/Directory.Build.props#L23C21-L23C56), langVersion 9 is the most up to date version!

keegan-caruso:
The langVersion that is available is based on the compiler that is installed not what you are targeting.

keegan-caruso:
need to update: https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/bcc6cc12dc491e62f438d19fc6d46f33dd2756e5/.github/workflows/dotnetcore.yml

test/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
 	<PropertyGroup>
+		<TargetFrameworks>net8.0</TargetFrameworks>
```
keegan-caruso, line 8: line 17 will never be evaluated with this change

kellyyangsong: Thanks for the call out, I've removed this line

keegan-caruso:
one more :)

https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/build/template-install-dotnet-core.yaml

test/Microsoft.Identity.Abstractions.Tests/Microsoft.Identity.Abstractions.Tests.csproj
kellyyangsong: I was able to make the appropriate adjustments!

keegan-caruso (approved):
thanks

jennyf19 (approved):
:shipit:

PR #118, opened 2024-02-16
Fixed invalid JSON example

JoshLozensky:
Added an empty value to JSON object in config example to fix what was otherwise a malformed JSON example.


jmprieur (approved):
LGTM
Thanks @JoshLozensky

PR #129, opened 2024-06-14
Introduces IAuthorizationHeaderProviderExtension

sruke:
What?
This PR updates the IAuthorizationHeaderProvider interface to include the CreateAuthorizationHeaderAsync API. This new API allows for acquiring an authorization header on behalf of either a user or an application. This is a breaking change. To accomodate it Microsoft.Identity.Abstractions will update its major version. 

Why?
The current interface exposes separate APIs for obtaining headers on behalf of a user and an app.
The existing method for acquiring a header on behalf of an app does not accept a ClaimsPrincipal, limiting the flexibility and usability of the current API.


src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Provides a method to create an authorization header for calling protected web APIs 
+    /// on behalf of a user or the application itself.
+    /// </summary>
+    public interface IAuthorizationHeaderProviderExtension
```
jennyf19, line 15: Let's remove "extension" as it's not exactly an extension.

src/Microsoft.Identity.Abstractions/ApplicationOptions/RequestContext.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.Identity.Abstractions
+{
+    /// <summary>
+    /// Holds information about an incoming request.
+    /// </summary>
+    public class RequestContext
```
jennyf19, line 13: not sure why we need this, can you give more details?

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
sruke: Makes sense. That can be done.

src/Microsoft.Identity.Abstractions/ApplicationOptions/RequestContext.cs
sruke: The PR description provides some detail on the usefulness of the RequestContext.

It is not needed today but offers flexibility for the future. The PropertyBag on the RequestContext can hold additional information about the request from the user or Downstream API, preventing the need for a new interface. This is similar to Wilson's CallContext, which is used throughout the call graph.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <param name="requestContext">Information about the incoming request.</param>
+        /// <param name="scopes">The scopes for which to request the authorization header.</param>
+        /// <param name="options">Options containing information about the API to be called and token acquisition settings.</param>
+        /// <param name="subject">The claims principal representing the authenticated user or entity making the request.
+        /// Typically provided in a web API or web app context; not usually needed in daemon applications.</param>
+        /// <param name="cancellationToken">A token to cancel the operation.</param>
+        /// <returns>A string containing the authorization header, such as "Bearer token" or "PoP token".</returns>
+        Task<string> CreateAuthorizationHeaderAsync(
+            RequestContext requestContext,
+            IEnumerable<string> scopes,
```
keegan-caruso, line 29: We tried this in Mise in earlier versions. It confused people to have a single method for app and user.

For example, when requesting a token for an app, we should only accept a single string for the scope, to have an IEnumerable is misleading.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
keegan-caruso, line 15: Generally, when following this pattern, this interface would also implement the base interface.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
```diff
@@ -49,5 +49,24 @@ public interface IAuthorizationHeaderProvider
...
+        /// Provide a single scope if the header needs to be created on behalf of an application.</param>
+        /// <param name="options">The <see cref="AuthorizationHeaderProviderOptions"/> containing information about the API
+        /// to be called and token acquisition settings. If not provided, the header will be for a bearer token.</param>
+        /// <param name="claimsPrincipal">Inbound authentication elements. In a web API, this is usually the result of the
+        /// validation of a token. In a web app, this would be information about the signed-in user. This is not useful in
+        /// daemon applications. In Microsoft.Identity.Web you rarely need to provide this parameter as it's inferred from the
+        /// context.</param>
+        /// <param name="cancellationToken">A token to cancel the operation.</param>
+        /// <returns>A string containing the authorization header, such as "Bearer token" or "PoP token".</returns>
+        Task<string> CreateAuthorizationHeaderAsync(
```
keegan-caruso, line 66: Isn't this a major version change?

sruke: Yes, Abstractions will increment the major version to accommodate the breaking change.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
sruke: There was a change in direction. The existing IAuthorizationHeaderProvider interface was updated instead.

src/Microsoft.Identity.Abstractions/ApplicationOptions/RequestContext.cs
sruke: Removed RequestContext as it is not needed today.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
sruke: Good call out. When acquiring a header on-behalf-of an application RequestAppToken must be set to true. I've also clarified in the summary that only single scope is needed for apps. 

```
DevEx:
await authorizationHeaderProvider.CreateAuthorizationHeaderAsync(
     new [] { "https://graph.microsoft.com/.default" },
     new AuthorizationHeaderProviderOptions { RequestAppToken = true }).ConfigureAwait(false);
```

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
```diff
@@ -49,5 +49,24 @@ public interface IAuthorizationHeaderProvider
...
+            IEnumerable<string> scopes,
```
brentschmaltz, line 67: Do we have to pass IEnumerable<string> can it be IList<string>?

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
```diff
@@ -49,5 +49,24 @@ public interface IAuthorizationHeaderProvider
...
+            AuthorizationHeaderProviderOptions? options = null,
+            ClaimsPrincipal? claimsPrincipal = null,
```
pmaytak, line 69: So a ClaimsPrincipal can be passed when acquiring a token for an app? What will it be used for in that case?

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
pmaytak: IEnumerable is consistent with the CreateAuthorizationHeaderForUserAsync signature. However, sometimes we manipulate the scopes (like adding default ones), in that case IList would be better. I prefer the IList, unless you'd prefer IEnumerable for consistency.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
pmaytak, line 1: nit: suggestion to clarify the comment on line 40 to something like this:
```csharp
/// <param name="scopes">Scope for which to request the authorization header (for example, "https://graph.microsoft.com/.default"</param>
```

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProviderExtension.cs
pmaytak: +1. I prefer to have separate methods since the parameters to pass in are also a bit different, so it seems like that's more clear on what flow we're doing.

jmprieur: @keegan-caruso @pmaytak thanks for your feedback.
The separate methods are still here. But you also know that IDownstreamApi has methods named CallApiAsync, GetApiAsync, PostApiAsync etc ... which are parameterized by just a service name (that is the configuration, in which the RequestAppToken is set or not).

In other words, what Sruthi is proposing is an even higher level API on IAuthorizationHeaderProvider to accomodate for this scenario, and other 1P specific protocols.

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
jmprieur: It's supposed to be immutable. No need to manipulate anything. IEnumerable<string> is the base interface for string[], IList<string> etc ...

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
jmprieur: Some 1P protocols where you have several tokens, @pmaytak

jmprieur (approved):
LGTM
thanks @sruke 

You might want to up the version in ?
https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/cade7c8a697aba33391ed3b373b6f1266f909c34/Directory.Build.props#L4

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider.cs
sruke: Scopes can be of many forms. There are all valid scopes:
- "api://634de702-3173-4a71-b336-a4fab786a479/.default", 
- "https://graph.microsoft.com/.default, 
- "user.read", 
- "api://634de702-3173-4a71-b336-a4fab786a479/access_as_user"

Scopes are also defined based on the header that needs to be acquired (is it on-behalf-of an app or user). ID.Web has documentation on the accepted forms of scopes for each scenario.

PR #131, opened 2024-07-19
IDownstreamApi overloads that take JsonTypeInfo<T> as a parameter to enable source generated Json deserialization for NativeAOT

dualtagh:
https://github.com/AzureAD/microsoft-identity-web/issues/2930

Adding overloads to IDownstreamApi that accept JsonTypeInfo<T> for inputs / outputs so that source generated Json serialization can work in a NativeAOT context.


Directory.Build.props
```diff
@@ -46,6 +46,10 @@
 	<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
 		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
 	</PropertyGroup>
+    
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
jmprieur, line 50: do we want to do it for net8.0 and net9.0?

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.tt
```diff
@@ -64,6 +73,13 @@ namespace Microsoft.Identity.Abstractions
 <# if (hasOutput){ #>
         /// <typeparam name="TOutput">Generic output type.</typeparam>
 <# } #>
+
+<# if (hasInput && framework == "net8"){ #>
+        /// <typeparam name="JsonTypeInfo<TInput>">JSON serialization metadata for TInput</typeparam>
+<# } #>
+<# if (hasOutput && framework == "net8"){ #>
+        /// <typeparam name="JsonTypeInfo<TOutput>">JSON serialization metadata for TOutput</typeparam>
+<# } #>
```
jmprieur, line 82: You might want to move the content of this line at the end of the previous one to avoid generating an extra empty line in the generated code.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.tt
```diff
@@ -52,7 +58,10 @@ namespace Microsoft.Identity.Abstractions
      string user = !hasApp ? "user" : "null";
      string sampleCodeTemplate = template.Replace("TOutput", "IEnumerable<MyItem>").Replace("TInput", "MyItem").Replace("<", "&lt;").Replace(">", "&gt;");
 #>
+<# if (framework == "net8"){ #>
```
jmprieur, line 61: You could actually move this line right after line 32, and avoid having a `#if` around each method in the generated code. 
Line 63 and 64 would have to go after line 147

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.cs
```diff
@@ -236,5 +235,136 @@ public partial interface IDownstreamApi
...
+        /// </code>
+        ///
+        /// Example of editing.
+        /// <code>
+        /// public Task&lt;MyItem&gt; EditAsync(MyItem myItem)
+        /// {
+        ///   return _downstreamWebApi.CallWebApiForUserAsync&lt;MyItem, MyItem&gt;(
+        ///         ServiceName,
+        ///         nyItem,
+        ///         options =>
```
jmprieur, line 281: You might want to update this code snippet (which goes in the reference documentation). See the Examples paragraph of https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.abstractions.idownstreamapi.callapiforuserasync?view=msal-model-dotnet-latest for instance

src/Microsoft.Identity.Abstractions/Microsoft.Identity.Abstractions.csproj
```diff
@@ -27,4 +27,7 @@
     </Compile>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
jmprieur, line 30: I would say net8.0 or net9.0

jmprieur (approved):
LGTM
I added a few suggestions

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.tt
dualtagh: I think I've done what you are asking

jmprieur: Thanks @dualtagh

jmprieur (approved):
LGTM
Thanks @dualtagh 

you might want to run:

```
dotnet pack /p:GenerateCompatibilitySuppressionFile=true
```

to generate the API compat suppression files (until we move to version 7.0, at which point you'd remove them?

Directory.Build.props
```diff
@@ -46,6 +46,10 @@
 	<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
 		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
 	</PropertyGroup>
+    
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0' or '$(TargetFramework)' == 'net9.0'">
```
eerhardt, line 50:
```suggestion
  <PropertyGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
```

This is a more maintainable way of writing this condition.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
```diff
@@ -765,5 +769,756 @@ public partial interface IDownstreamApi
...
+        /// </example>
+        public Task<TOutput?> PutForAppAsync<TInput, TOutput>(
+            string? serviceName,
+            TInput input,
+            JsonTypeInfo<TInput> inputJsonTypeInfo,
+            JsonTypeInfo<TOutput> outputJsonTypeInfo,
+            Action<DownstreamApiOptionsReadOnlyHttpMethod>? downstreamApiOptionsOverride = null,
+            CancellationToken cancellationToken = default)where TOutput : class;
+
+#if NETSTANDARD2_1_OR_GREATER
```
eerhardt, line 1219: I'm not sure this #if makes sense`. You are already in a "if .NET 8 or greater" condition. This code won't be used in a netstandard2.1 compilation.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.cs
```diff
@@ -236,5 +240,136 @@ public partial interface IDownstreamApi
...
+        ///         {
+        ///            options.HttpMethod = HttpMethod.Patch;
+        ///            options.RelativePath = $"api/todolist/{myItem.Id}";
+        ///         });
+        /// }
+        /// </code>
+        /// </example>
+        public Task<TOutput?> CallApiForUserAsync<TInput, TOutput>(
```
eerhardt, line 294: Are we concerned at all about adding methods to an existing interface? We typically treat that as a breaking change in .NET. Would it make more sense to add a new interface that callers can cast to?

dualtagh: Discussed with @keegan-caruso and we agreed to break the interface and bump the major version which is pattern the team used in the past

eerhardt: Sounds good to me.

Directory.Build.props
```diff
@@ -47,6 +47,10 @@
 		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
 	</PropertyGroup>
 
+  <PropertyGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
```
eerhardt, line 50: I'm a bit confused. Are you trying to say "if you don't target net8.0, then use the 8.0.4 System.Text.Json package"?

Note that you don't need to explicitly reference the System.Text.Json package if you target `net8.0`. It is available automatically.

dualtagh: Sorry - typo for the `!` - I wanted to include System.Text.Json for net8+.
As you mentioned offline, I can reference System.Text.Json for all the target frameworks so I've updated to that.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.cs
```diff
@@ -236,5 +237,136 @@ public partial interface IDownstreamApi
             Action<DownstreamApiOptions>? downstreamApiOptionsOverride = null,
             CancellationToken cancellationToken = default)
             where TOutput : class;
+
+#if NET8_0_OR_GREATER
```
eerhardt, line 241: Since you are always referencing System.Text.Json on all TFMs, you don't need to only put these APIs in for `net8`. You can remove the `#if` and always compile them in.

That also means you don't need an explicit target for `net8` since all targets will have these new methods.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.cs
```diff
@@ -236,5 +237,136 @@ public partial interface IDownstreamApi
             Action<DownstreamApiOptions>? downstreamApiOptionsOverride = null,
```
eerhardt, line 237: The above has

```
#if NET6_0_OR_GREATER
        [RequiresUnreferencedCode("This method's implementations also use generic types and are not trim-friendly.")]
#endif
```

However no one was building for net6+ before. In order to make this attribute be compiled, you will need at least one TFM that is `net6.0` or greater.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.cs
dualtagh: Switched back to net8+ only to avoid breaking existing users

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.cs
dualtagh: This is now removed inside the net8 block

Directory.Build.props
dualtagh: Thanks, this is now removed

jennyf19: @dualtagh we probably have similar issues in all our repos. Can you either fix them all or open issues? thank you.

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Abstractions/CompatibilitySuppressions.xml
```diff
@@ -1,6 +1,30 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids -->
 <Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
```
eerhardt, line 4: Are these still necessary?

eerhardt: Oh, I think they are because you aren't putting the Patch APIs in the net8 target.

We should fix the .tt file to also include those for net8.

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.tt
```diff
@@ -24,9 +28,20 @@ namespace Microsoft.Identity.Abstractions
...
+#>
+
+#if NET8_0_OR_GREATER
+<#
+    }
+
   foreach(string httpMethod in new string[]{  "Get", "Post", "Put", "Patch", "Delete"} )
   {
-   if (httpMethod == "Patch")
+   if (httpMethod == "Patch" && framework != "net8")
```
eerhardt, line 44: I think you still want these "Patch" methods on net8. We probably need to update line 48 to:

```
#if NETSTANDARD2_1_OR_GREATER || NET
```

or similar.

src/Microsoft.Identity.Abstractions/CompatibilitySuppressions.xml
dualtagh: Thanks for catching that - I've updated the tt file

dualtagh: The remaining suppresions are the new JsonTypeInfo interfaces

src/Microsoft.Identity.Abstractions/DownstreamApi/IDownstreamApi.HttpMethods.tt
```diff
@@ -24,13 +28,24 @@ namespace Microsoft.Identity.Abstractions
...
+    {
+#>
+
+#if NET8_0_OR_GREATER
+<#
+    }
+
   foreach(string httpMethod in new string[]{  "Get", "Post", "Put", "Patch", "Delete"} )
   {
    if (httpMethod == "Patch")
```
eerhardt, line 44: (nit) 
```suggestion
   if (httpMethod == "Patch" && framework != "net8")
```

This will remove the nested `#if NETSTANDARD2_1_OR_GREATER || NET` when you are already in the `#if NET8_0_OR_GREATER` condition. That nested `#if` is unnecessary since you are already in .NET 8+.

eerhardt (approved):
LGTM. Nice work.

PR #134, opened 2024-08-12
[test] updates for one build

jennyf19:
Ensures that the same build variable name (`MicrosoftIdentityAbstractionsVersion`) is used both for the version of the library itself, and in the libraries that take a dependency on Wilson. This name was already used in other libraries. This becomes the way to set the version in Release builds.


bgavrilMS (approved):

jmprieur (approved):
LGTM
Thanks @jennyf19

PR #148, opened 2025-01-21
Fix #147

### LINKED ISSUE #147: [Bug] `Id` property in `CredentialDescription` is derived from secret values
Refer to AzureAD/microsoft-identity-web#3201 (no longer available)

Problematic code:
https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/38dbb1db33b7af75fc892705bedf999efa1f5fcf/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs#L343-L350

jmprieur:
# Fix #147

- Fixes #147. This is a desired behavioral change when reading the "ReferenceOrValue" in the case of ClientSecret.
- Updates the public API (that should have been updated earlier)
- Update the library version number to the next version.

This does not require changes to the consuming libraries. Unified build (20250121.2) started to confirm.

### DIFF
--- a/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
+++ b/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
@@ -347,7 +347,7 @@ namespace Microsoft.Identity.Abstractions
                     CredentialSource.StoreWithDistinguishedName => CertificateDistinguishedName,
                     CredentialSource.Certificate or CredentialSource.Base64Encoded => Base64EncodedValue,
                     CredentialSource.SignedAssertionFromManagedIdentity => ManagedIdentityClientId,
-                    CredentialSource.ClientSecret => ClientSecret,
+                    CredentialSource.ClientSecret => "***",
                     _ => null,
                 };
             }
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
         <!-- This should be passed from the VSTS build -->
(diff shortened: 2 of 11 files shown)


jennyf19 (approved):

pmaytak (approved):

bgavrilMS (approved):

kllysng (approved):

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -347,7 +347,7 @@ public string? ReferenceOrValue
                     CredentialSource.StoreWithDistinguishedName => CertificateDistinguishedName,
                     CredentialSource.Certificate or CredentialSource.Base64Encoded => Base64EncodedValue,
                     CredentialSource.SignedAssertionFromManagedIdentity => ManagedIdentityClientId,
-                    CredentialSource.ClientSecret => ClientSecret,
+                    CredentialSource.ClientSecret => "***",
```
msbw2, line 350: @jmprieur The comment on the `Id` property about the value being a `unique identifier` is no longer accurate with this change since all `ClientSecret` sources will have the same `Id`. This could negatively impact usages of this `Id` where the value is expected to be unique to that specific source/value.

bgavrilMS: What is the scenario where you'd use 2 client secrets?

msbw2: I don't have a specific scenario in mind. The main point is that the documentation indicates that the `Id` is unique which is outdated or, at the very least, misleading.

PR #151, opened 2025-01-23
Removing the Container and ValueOrReference from the public API of CredentialDescription

jmprieur:
Removing the Container and ValueOrReference from the public API of CredentialDescription. They were there for compat with IdWeb 1.x and part of the technical debt

# Removing the Container and ValueOrReference from the public API of CredentialDescription
- Removing technical debt. These properties are now internal, which avoids them to be serialized or bound, which is no longer desirable since Id.Web 2.x
- This is a breaking change (these properties become internal) so bumping the library to 8.0.0.


src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -15,6 +16,45 @@ namespace Microsoft.Identity.Abstractions
...
+        /// </summary>
+        public CredentialDescription()
+        {
+
+        }
+
+        /// <summary>
+        /// Copy constructor for <see cref="CredentialDescription"/>
+        /// </summary>
+        /// <param name="other"></param>
```
kllysng, line 30: Non-blocking: consider removing this or adding a description

kllysng (approved):
LGTM

JoshLozensky (approved):

PR #166, opened 2025-03-24
Compute really unique ID for ClientCredentials

jmprieur:
# Fix163

## Description

- Removes the container and value properties that are no longer needed since IdWeb 2
- Compute the ID directly, based on the credential type, making sensitive data is redacted

### DIFF
--- a/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
+++ b/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Buffers.Text;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
@@ -82,6 +83,10 @@ namespace Microsoft.Identity.Abstractions
                             {
                                 _cachedId = $"Certificate={Certificate.Thumbprint}";
                             }
+                            else
+                            {
+                                _cachedId = $"Certificate=null";
+                            }
                             break;
                         case CredentialSource.KeyVault:
                             _cachedId = $"CertificateFromKeyVault={KeyVaultUrl}/{KeyVaultCertificateName}";
@@ -135,7 +140,7 @@ namespace Microsoft.Identity.Abstractions
 #if !NET8_0_OR_GREATER
             using (SHA256 sha256 = SHA256.Create())
             {
-                digest = sha256.ComputeHash(Encoding.UTF8.GetBytes(secret));
+                digest = sha256.ComputeHash(Encoding.Unicode.GetBytes(secret));
             }
 #else
             digest = SHA256.HashData(Encoding.Unicode.GetBytes(secret));
--- a/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
+++ b/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
@@ -3,8 +3,12 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -63,17 +67,80 @@ public CredentialDescription(CredentialDescription other)
...
+                            break;
+                        case CredentialSource.CustomSignedAssertion:
+                            string parameterNames = string.Join(",", (IEnumerable<string>?)CustomSignedAssertionProviderData?.Keys ?? []);
+                            _cachedId = $"CustomSignedAssertion={CustomSignedAssertionProviderName}({parameterNames})";
+                            break;
+                        default:
+                            throw new ArgumentException("Unknown credential source type");
+                    }
+                }
+                return _cachedId!;
```
keegan-caruso, line 124: This could be null if the source type is cert, and the cert is null.

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -63,17 +67,80 @@ public CredentialDescription(CredentialDescription other)
...
-                return _cachedId;
+#if !NET8_0_OR_GREATER
+            using (SHA256 sha256 = SHA256.Create())
+            {
+                digest = sha256.ComputeHash(Encoding.UTF8.GetBytes(secret));
             }
+#else
+            digest = SHA256.HashData(Encoding.Unicode.GetBytes(secret));
+#endif
+            return Convert.ToBase64String(digest);
```
keegan-caruso, line 143: minor, but System.Text.Buffers.Base64 is preferred for new code.

test/Microsoft.Identity.Abstractions.Tests/CredentialDescriptionIdTest.cs
```diff
@@ -0,0 +1,185 @@
+using Microsoft.Identity.Abstractions;
```
keegan-caruso, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

using Microsoft.Identity.Abstractions;
```

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
keegan-caruso: Ah, never mind, this is a pretty optimized method.

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
keegan-caruso: This gets used as a cache key in IdWeb, so this would throw if null and used there. Would be better to throw here if we can't create the id.

jmprieur: I'l make it non-null. as I can see cases where the cert would be set afterwards (some customers want to do that)

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -2,9 +2,14 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Buffers.Text;
```
keegan-caruso, line 5:
```suggestion
```

keegan-caruso (approved):
Small comment, otherwise LGTM

jennyf19 (approved):

PR #177, opened 2025-04-22
Adding serializer for CredentialDescription in .NET 8+

### LINKED ISSUE #176: [Feature Request] Add a custom JSon serializer for CredentialDescription in .NET 8+
**Is your feature request related to a problem? Please describe.**
When serializing CredentialsDescription instances, there is a lot of information that is not relevant. It would be good to have a Custom json serializer at least in .NET 8 (to avoid brining new dependencies in .NET 462, netstandard) 

**Describe the solution you'd like**
Add a custom JSon serializer. the Write method would only write the right properties based on the value of the sourceType enumeration. Do not write the Id, do not write the Cache, the Certificate and Skip

**Describe alternatives you've considered**
Add to all target framwork, but it's really needed for tools (written in .NET 8+)

jmprieur:
# Add a custom serializer for CredentialDescription in .NET 8.

Fixes #176

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
         <!-- This should be passed from the VSTS build -->
-        <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">9.1.0</MicrosoftIdentityAbstractionsVersion>
+        <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">9.2.0</MicrosoftIdentityAbstractionsVersion>
         <!-- This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion -->
         <Version>$(MicrosoftIdentityAbstractionsVersion)</Version>
         <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)\build\35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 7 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
         <!-- This should be passed from the VSTS build -->
-        <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">9.1.0</MicrosoftIdentityAbstractionsVersion>
+        <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">9.2.0</MicrosoftIdentityAbstractionsVersion>
```
keegan-caruso, line 4:
```suggestion
        <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">9.1.0</MicrosoftIdentityAbstractionsVersion>
```

This should stay at 9.1.0, 9.1.0 hasn't been released yet

keegan-caruso (approved):
Other than the version, it looks right

Directory.Build.props
jmprieur: yes. I realized yesterday night. Thanks!

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescriptionJsonConverter.cs
```diff
@@ -0,0 +1,196 @@
...
+                {
+                    throw new JsonException("Expected property name");
+                }
+
+                string propertyName = reader.GetString()!;
+                reader.Read();
+
+                switch (propertyName.ToLowerInvariant())
+                {
+                    case "sourcetype":
```
keegan-caruso, line 46: could we do 

```cs
case nameof(CredentialDescription.SourceType)
```

And similar for other values?

keegan-caruso: Or not worth it to remove the string literals?

jmprieur: Thanks for the suggestion, @keegan-caruso . I did it!

jennyf19 (approved):

PR #183, opened 2025-05-14
Add Algorithm property to CredentialDescription

RojaEnnam:
## Description

Add Algorithm property to CredentialDescription


src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -145,6 +145,11 @@ private string GenerateHash(string? secret)
             return Convert.ToBase64String(digest);
         }
 
+        /// <summary>
+        /// Algorithm used to create signing credentials.
+        /// </summary>
+        public string Algorithm { get; set; }
```
keegan-caruso, line 151:
```suggestion
        public string? Algorithm { get; set; }
```

JoshLozensky:
Looks like it needs to be added to the api docs

jmprieur (changes requested):
Thanks @RojaEnnam
Please update the unshippes public Api (using the fixer in Visual Studio), and update the xml comment to be lilenthe others properties. Add unit tests. You can use an agent to help you

Abstraction currently has 100% test coverage and the tests are used in the reference documentation

RojaEnnam:
Updated the unshipped public api and added unit tests.

jmprieur:
> Updated the unshipped public api and added unit tests.

You only updated it for one of the target framework?

test/Microsoft.Identity.Abstractions.Tests/CredentialDescriptionTest.cs
```diff
@@ -461,5 +454,37 @@ public void CopyConstructorFromNull_ShouldThrow()
             _ = Assert.Throws<ArgumentNullException>(() => new CredentialDescription(original!));
 #pragma warning restore CS8600 // Converting null literal or possible null value to non-nullable type.
         }
+
+        [Fact]
+        public void Algorithm_SetAndGet_ShouldWork()
+        {
```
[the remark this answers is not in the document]
RojaEnnam: Added the comment

JoshLozensky:
FYI @RojaEnnam in VisualStudio if you show potential fixes or do quick actions and refactoring for the missing API (the set method in this case) you can have the IDE do this for you. If you choose the 'fix all occurrences in: document' option it will add it to all relevant API files in one go.

image

JoshLozensky (approved):

jmprieur:
> FYI @RojaEnnam in VisualStudio if you show potential fixes or do quick actions and refactoring for the missing API (the set method in this case) you can have the IDE do this for you. If you choose the 'fix all occurrences in: document' option it will add it to all relevant API files in one go.
> 
> image

You might have to do if for each target framework (choosing the target framework in the combo-box)

test/Microsoft.Identity.Abstractions.Tests/CredentialDescriptionTest.cs
```diff
@@ -461,5 +454,50 @@ public void CopyConstructorFromNull_ShouldThrow()
...
+            /*
+            // <algorithm_json>
+            {
+                "ClientCredentials": [
+                {
+                    "Algorithm": "RS256",
```
jmprieur, line 466: Still not right. This only works with a certificate, @RojaEnnam
you need to take one of the cases with a certificate (for instance KeyVault)

jmprieur (approved):
LGTM
please update the XML comment (and the test) with an example where the cert is coming from KeyVault to show how this will be used.

test/Microsoft.Identity.Abstractions.Tests/CredentialDescriptionTest.cs
RojaEnnam: Addressed. Thanks.

JoshLozensky (approved):

PR #196, opened 2025-06-25
Remove class restriction on OperationResult

brentschmaltz:
Add implicit operator
Disable empty ctor.

### DIFF
--- a/src/Microsoft.Identity.Abstractions/Results/OperationResult.cs
+++ b/src/Microsoft.Identity.Abstractions/Results/OperationResult.cs
@@ -6,6 +6,8 @@ using System.Diagnostics.CodeAnalysis;
 #endif
 
 #nullable enable
+using System;
+
 namespace Microsoft.Identity.Abstractions
 {
     /// <summary>
@@ -19,7 +21,6 @@ namespace Microsoft.Identity.Abstractions
     /// This pattern helps with clear error handling and propagation.
     /// </remarks>
     public readonly record struct OperationResult<TResult, TError>
-            where TResult : class
             where TError : OperationError
     {
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Microsoft.Identity.Abstractions/Results/OperationResult.cs
```diff
@@ -72,6 +81,28 @@ public OperationResult(TError error)
         /// <returns>The result associated with the operation result.</returns>
         /// <remarks>This property is only valid if the result type is valid.</remarks>
         public readonly TResult? Result { get; }
+
+#pragma warning disable RS0036 // Annotate nullability of public types and members in the declared API
```
keegan-caruso, line 85: We shouldn't suppress this one, these API files are all using nullables

src/Microsoft.Identity.Abstractions/Results/OperationResult.cs
```diff
@@ -19,8 +21,7 @@ namespace Microsoft.Identity.Abstractions
     /// This pattern helps with clear error handling and propagation.
     /// </remarks>
     public readonly record struct OperationResult<TResult, TError>
-            where TResult : class
-            where TError : OperationError
+        where TError : OperationError
```
pmaytak, line 24: Can you add more tests to https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/test/Microsoft.Identity.Abstractions.Tests/OperationResultTests.cs that use primitive types?

src/Microsoft.Identity.Abstractions/Results/OperationResult.cs
```diff
@@ -72,6 +81,25 @@ public OperationResult(TError error)
...
+        /// </summary>
+        /// <param name="error">The error to be stored in the result.</param>
+        public static implicit operator OperationResult<TResult, TError>(TError error) => new(error);
+
+        /// <summary>
+        /// Casts the result to a <see cref="OperationResult{TResult, TError}"/>.
+        /// </summary>
+        /// <remarks>Required for compatibility, see CA2225 for more information</remarks>
+        /// <returns>The existing instance.</returns>
+        public OperationResult<TResult, TError> ToOperationResult() => this;
```
pmaytak, line 102: What is the purpose/usage of this?

PR #212, opened 2025-10-07
Add Thumbprint to all ID fields

tlupes:
# Add Thumbprint to all ID fields

Add Thumbprint to all ID fields where applicable.

## Description

Microsoft.Identity.Web uses the ID field to determine caching of the confidential builder. If an ID does not change, the old certificate will continue to be used. As a result, some certificate types cannot be rotated as the old value will always be used. This change adjusts the ID field to always include the thumbprint to ensure that a new certificate will be immediately used.


keegan-caruso (approved):
Changes look correct, would be good to add a test

PR #252, opened 2026-05-13
Add CredentialDescription.UseBoundCredential for Bearer-over-mTLS opt-in

gladjohn:
## Summary
Adds a new `UseBoundCredential` (bool, default `false`) property on `CredentialDescription`. When set, the credential is treated as opted-in to the bound-credential (Bearer-over-mTLS) flow. The setting is per-credential: each `CredentialDescription` in `ClientCredentials[]` decides for itself, so an app can mix bound and non-bound credentials freely.

Consumers (MSAL.NET / Microsoft.Identity.Web) read this property to decide whether to:
- Present a certificate over mTLS (`CertificateOptions.SendCertificateOverMtls = true`) for `Certificate`-type credentials, or
- Wire a bound-credential bundle (signed assertion + binding certificate) via `WithClientAssertion(Func<…, ClientSignedAssertion>)` for `SignedAssertionFromManagedIdentity`-type credentials.

Companion to MSAL.NET issue: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5791

## Configuration shape
```json
{
  "ClientCredentials": [
    {
      "SourceType": "SignedAssertionFromManagedIdentity",
      "ManagedIdentityClientId": "your-user-assigned-managed-identity-client-id",
      "TokenExchangeUrl": "api://AzureADTokenExchange",
      "UseBoundCredential": true
    }
  ]
}
```

## Changes
- `CredentialDescription.UseBoundCredential` property + copy constructor propagation.
- `CredentialDescriptionJsonConverter` updated to round-trip the new property through System.Text.Json.
- `README.md` mermaid diagram updated.
- Public API tracker updated for all 6 TFMs (net462 / netstandard2.0 / netstandard2.1 / net8.0 / net9.0 / net10.0).
- Tests: default value, copy-constructor propagation, JSON round-trip via the converter.

## Validation
- Full unit test suite: **291 / 291 pass** (net8.0 + net10.0 + net462).
- Public API analyzer clean across all TFMs.

## Notes for reviewers
- The new property is purely declarative on the abstractions surface — no behavior change in this repo. The consumer libraries (MSAL.NET, Microsoft.Identity.Web) interpret it at credential-load time.
- Default is `false`, so existing configurations are unaffected.

### DIFF
--- a/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
+++ b/src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Identity.Abstractions
 #if NET8_0_OR_GREATER
             ArgumentNullException.ThrowIfNull(other);
 #else
-            if (other == null)
+            if (other is null)
                 throw new ArgumentNullException(nameof(other));
 #endif
             Algorithm = other.Algorithm;
@@ -63,6 +63,7 @@ namespace Microsoft.Identity.Abstractions
             SourceType = other.SourceType;
             TokenExchangeUrl = other.TokenExchangeUrl;
             TokenExchangeAuthority = other.TokenExchangeAuthority;
+            UseBoundCredential = other.UseBoundCredential;
         }
 
         private string? _cachedId;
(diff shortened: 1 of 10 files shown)


gladjohn:
@bgavrilMS @trwalke this is needed for this work - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5791 

IdWeb PR is blocked because of multiple issues (STJ + WithExtraBodyParameters) related to MSAL 4.84.0

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
```diff
@@ -572,6 +573,43 @@ public virtual object? CachedValue
...
+        ///         "TokenExchangeUrl": "api://AzureADTokenExchange",
+        ///         "UseBoundCredential": true
+        ///       }
+        ///     ]
+        ///   }
+        /// }
+        /// ```
+        /// ]]></format>
+        /// </example>
+        public bool UseBoundCredential { get; set; }
```
bgavrilMS, line 611: I think eventually we will want to change the default to be bound credentials. Should we rename this as "PreferMtlsCredential" and the client does a best effort?

gladjohn: @MZOLN thoughts on this

cpp11nullptr: Perhaps, `IsBoundCredential` would have a better semantic, since it's declarative.

cpp11nullptr (approved):

src/Microsoft.Identity.Abstractions/ApplicationOptions/CredentialDescription.cs
gladjohn: ## On the naming: let's pick the behavior first

Quick note since the name depends on what we actually want this to do.

Today this is a `bool`, default `false`, **best-effort**: use a bound credential where the host supports it, silently ignore it elsewhere. The name `UseBoundCredential` sounds stricter than that, which is the source of the confusion.

The real question is whether fallback is allowed:
- **Best-effort** (try to bind, fall back if unsupported) — what the code does now.
- **Strict** (bind or fail) — what we may want on a security-sensitive path.

**Based on offline conversations - votes exist for:** use a `BoundCredentialMode` enum (`Default`, `Disabled`, `Preferred`, `Required`) instead of a bool.
- `Required` covers the strict case.
- `Preferred` covers today's best-effort case.
- `Default` lets us change the default later without a breaking change.

Since the API is still unshipped, now's the cheap time to do it.

If we'd rather keep a bool, let's keep `UseBoundCredential` but update the doc to make the best-effort behavior explicit.

Please vote on: (1) best-effort vs strict, and (2) bool vs enum.

bgavrilMS (approved):

PR #262, opened 2026-07-13
Add OnBeforeAuthHeaderCreation / OnAfterAuthHeaderCreation hooks

neha-bhargava:
## What

Adds two `Action<HttpRequestMessage>?` hooks to `AuthorizationHeaderProviderOptions`, giving callers explicit, self-documenting control over request customization relative to authorization-header creation:

- **`OnBeforeAuthHeaderCreation`** — runs **before** the header is created. The `Authorization` header is not yet present. Use it to shape the request that a request-binding protocol signs — e.g. a SignedHttpRequest that binds the query, headers, or body (`q`/`h`/`b` claims) to the token — so the signature covers the finalized request.
- **`OnAfterAuthHeaderCreation`** — runs **after** the header (including `Authorization`) is set, just before the message is sent. Use it to observe or adjust the finalized request. Same timing as the existing `CustomizeHttpRequestMessage`.

## Why

`CustomizeHttpRequestMessage` is documented to run *after* the header, but the request-binding (SHR `q`/`h`/`b`) work needs a *before* hook. Rather than overload one callback with ambiguous timing, this adds an explicit before/after pair; the timing is encoded in the name.

## Notes

- Additive and non-breaking; existing `CustomizeHttpRequestMessage` behavior is unchanged.
- `PublicAPI.Unshipped.txt` updated across all 6 TFMs; version bumped to 12.5.0.
- Consumed by Microsoft.Identity.Web (`DownstreamApi`) in a companion PR.

### DIFF
--- a/src/Microsoft.Identity.Abstractions/DownstreamApi/AuthorizationHeaderProviderOptions.cs
+++ b/src/Microsoft.Identity.Abstractions/DownstreamApi/AuthorizationHeaderProviderOptions.cs
@@ -38,6 +38,8 @@ namespace Microsoft.Identity.Abstractions
             AcquireTokenOptions = other.AcquireTokenOptions.Clone();
             HttpMethod = other.HttpMethod.ToString();
             CustomizeHttpRequestMessage = other.CustomizeHttpRequestMessage;
+            OnBeforeAuthHeaderCreation = other.OnBeforeAuthHeaderCreation;
+            OnAfterAuthHeaderCreation = other.OnAfterAuthHeaderCreation;
             ProtocolScheme = other.ProtocolScheme;
             RequestAppToken = other.RequestAppToken;
         }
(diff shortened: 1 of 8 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- This should be passed from the VSTS build -->
-    <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">12.4.1</MicrosoftIdentityAbstractionsVersion>
+    <MicrosoftIdentityAbstractionsVersion Condition="'$(MicrosoftIdentityAbstractionsVersion)' == ''">12.5.0</MicrosoftIdentityAbstractionsVersion>
```
Avery-Dunn: Why combine the version bumps/changelog updates with source code changes? 

Aren't those usually handled after the release? https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/pull/260

src/Microsoft.Identity.Abstractions/DownstreamApi/AuthorizationHeaderProviderOptions.cs
```diff
@@ -75,6 +77,27 @@ public string HttpMethod
...
+        /// signature covers the finalized request. To observe or adjust the request <b>after</b> the header has been
+        /// created, use <see cref="OnAfterAuthHeaderCreation"/> (or <see cref="CustomizeHttpRequestMessage"/>).
+        /// </summary>
+        public Action<HttpRequestMessage>? OnBeforeAuthHeaderCreation { get; set; }
+
+        /// <summary>
+        /// Provides an opportunity for the caller app to customize the <see cref="HttpRequestMessage"/> <b>after</b>
+        /// the authorization header (including the <c>Authorization</c> header) has been set, and just before the
+        /// message is sent — the point at which the finalized request can be read or adjusted. Provided for naming
+        /// symmetry with <see cref="OnBeforeAuthHeaderCreation"/>; it shares this timing with the pre-existing
```
Avery-Dunn, line 94: >it shares this timing with the pre-existing...

The two hooks share timing, but does ordering matter? If it does then I suppose that's a problem for ID Web to handle, but it might be good to include any recommendations about ordering here in the comment to make it less ambiguous.

Avery-Dunn (approved):

Directory.Build.props
neha-bhargava: Good point - pulled the version bump and changelog out; they'll go in the post-release PR.

src/Microsoft.Identity.Abstractions/DownstreamApi/AuthorizationHeaderProviderOptions.cs
neha-bhargava: Added a note: they run at the same point, so callers shouldn't depend on their relative order.

Avery-Dunn (approved):

gladjohn (approved):

PR #266, opened 2026-08-04
Add `ICloudMetadataProvider` neutral contract for cross-cloud metadata

Avery-Dunn:
Introduces the neutral, dependency-free contract that lets one SDK contribute cloud-specific metadata and
another consume it without either depending on the other's concrete types. Adds the `ICloudMetadataProvider`
interface, the well-known `CloudMetadataKeyNames` vocabulary, and a ready-made mutable
`InMemoryCloudMetadataProvider`. The package ships **no cloud data** — it is only the contract shape.

Related work in MSAL and ID Web:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6104
https://github.com/AzureAD/microsoft-identity-web/pull/3994

## Motivation

Cloud-specific values (starting with the FIC token-exchange audience, e.g. `api://AzureADTokenExchangeUSGov`)
are owned by two different SDKs: the **public** list lives in MSAL and the **internal-only** list lives in
MISE. Microsoft.Identity.Web sits between them and must let an upstream-supplied provider flow through to its
own resolution without ID Web taking a hard dependency on the upstream SDK's types (or vice-versa). MSAL
cannot reference `Microsoft.Identity.Abstractions` (it references Wilson's `Microsoft.IdentityModel.Abstractions`),
so the two string bags cannot be unified into one type.

This PR adds the neutral contract + DI seam in `Microsoft.Identity.Abstractions` so any layer can register a
provider and any other layer can resolve it from the shared container, keyed off the authority host, over a
vocabulary of string keys that is identical across the SDKs (zero-remap translation pivot).

## Design

The surface is intentionally small — a key vocabulary, one interface, and one ready-made implementation, all
in `Microsoft.Identity.Abstractions`. Cloud metadata is represented directly as a plain
`IReadOnlyDictionary<string, string>` (no wrapper DTO), so nothing new has to be learned to read it.

### `CloudMetadataKeyNames` — the key vocabulary

```csharp
public static class CloudMetadataKeyNames
{
    public const string FederatedCredentialAudience = "federated_credential_audience"; // stored bare, no "/.default"
}
```

Values are addressed by well-known key constants rather than fixed typed properties, so new values can be
added — and obsolete ones retired (`[Obsolete]` then removed) — without source- or binary-breaking callers
(an unknown/removed key simply misses and the consumer falls back to a documented default). The literal is
intentionally **identical** to MSAL's `CloudMetadataKeyNames.FederatedCredentialAudience` and MISE's
equivalent, so a value written under a key by one layer is read under the same key by another with no
translation table.

### `ICloudMetadataProvider` — the resolution + DI seam

```csharp
public interface ICloudMetadataProvider
{
    IReadOnlyDictionary<string, string>? GetByAuthorityHost(string authorityHost); // null if the host is unknown
}
```

This is the seam registered in DI: e.g. MISE registers an implementation that knows internal-only sovereign
clouds, and Microsoft.Identity.Web resolves `ICloudMetadataProvider` from the container to obtain the correct
FIC audience for those clouds. Returning `null` for an unknown host lets a consumer fall back to another
provider (or a default). The return is a read-only, case-insensitive dictionary keyed by `CloudMetadataKeyNames`.

### `InMemoryCloudMetadataProvider` — a ready-made mutable provider

```csharp
// Inject/adjust a cloud (per-cloud or per-key); resolution layers over the optional fallback per key (own wins):
ICloudMetadataProvider provider = new InMemoryCloudMetadataProvider(fallback: upstreamProvider)
    .AddOrUpdate("login.mynewcloud.example", new Dictionary<string, string>
    {
        [CloudMetadataKeyNames.FederatedCredentialAudience] = "api://AzureADTokenExchangeMyCloud",
    })
    .AddOrUpdate("login.microsoftonline.us", CloudMetadataKeyNames.FederatedCredentialAudience, "api://Custom");

services.AddSingleton<ICloudMetadataProvider>(provider);
```

For callers who need to inject a new cloud or adjust an existing one without hand-implementing the interface,
with an optional fallback provider. Resolution merges the registered values over the fallback **per key**: a
key registered here wins, a key not registered here falls back to the same host in the fallback, and a host
unknown to both resolves to `null`. So a caller can adjust a single field of one cloud while inheriting the
rest.

It follows a **configure-then-read** model that matches the rest of the identity-SDK config surface: it is
populated at startup (the `AddOrUpdate` mutators are not thread-safe), then registered as a singleton and read
concurrently (reads are safe once population is complete). It is backed by a plain case-insensitive
`Dictionary`, not a concurrent collection — consistent with how MSAL builds its configuration.

### Key design decisions

1. **Contract only, no data.** The package supplies the neutral interface + vocabulary + ready-made provider;
   the public cloud list is owned by MSAL and the internal-only list by MISE. This keeps the SDKs decoupled
   while sharing one shape.
2. **Plain dictionary, no wrapper DTO.** Metadata is a `IReadOnlyDictionary<string, string>` rather than a
   dedicated bag class, so the contract stays minimal and mirrors MSAL's `KnownCloudMetadata.GetByAuthorityHost`
   return type exactly (the two dependency-separated stacks line up 1:1).
3. **Extensible, add/remove-safe surface.** Keys are `const` literals in `CloudMetadataKeyNames`; they can be
   added without breaking callers, and a key that must be retired is marked `[Obsolete]` for at least one
   release before removal.
4. **Method name aligned with MSAL.** `GetByAuthorityHost` is the canonical cross-SDK name — identical to
   MSAL's `KnownCloudMetadata.GetByAuthorityHost` — so the neutral contract and the MSAL baseline are drop-in
   parallel for ID Web's layered resolver.

## Downstream impact (companion PRs)

This PR is the neutral contract the other SDKs build on:

- **Microsoft.Identity.Web**: resolves `ICloudMetadataProvider` from DI (upstream/MISE-supplied), layered over
  MSAL's public baseline; both expose the audience under the identical key literal, so ID Web "translates" by
  reading the same key from whichever source resolves first.
- **MISE**: registers an `ICloudMetadataProvider` implementation carrying its internal-only clouds/metadata,
  which ID Web composes over MSAL's baseline.

## Public API surface

Namespace `Microsoft.Identity.Abstractions` (identical across all 6 TFMs — `net10.0` / `net9.0` / `net8.0` /
`net462` / `netstandard2.0` / `netstandard2.1`):

| API | Kind |
|---|---|
| `CloudMetadataKeyNames.FederatedCredentialAudience = "federated_credential_audience"` | static const |
| `ICloudMetadataProvider.GetByAuthorityHost(string) : IReadOnlyDictionary<string,string>?` | interface |
| `InMemoryCloudMetadataProvider(fallback = null)` + `AddOrUpdate(host, values)` + `AddOrUpdate(host, key, value)` + `GetByAuthorityHost` | sealed class |

## Changes

| File | Description |
|------|-------------|
| `CloudMetadata/CloudMetadataKeyNames.cs` | New public well-known key constants (`FederatedCredentialAudience`) |
| `CloudMetadata/ICloudMetadataProvider.cs` | New public interface — the DI seam; resolve metadata (a value dictionary) by authority host |
| `CloudMetadata/InMemoryCloudMetadataProvider.cs` | New ready-made mutable provider (`AddOrUpdate` ×2, optional fallback, per-key merge, configure-then-read) |
| `PublicAPI/*/PublicAPI.Unshipped.txt` (×6) | New public API surface entries |
| `CloudMetadataTests.cs` | Unit tests for the in-memory provider (resolution, merge, validation) |

## Test coverage

`CloudMetadataTests` (unit, 12 cases):
- `InMemoryCloudMetadataProvider` resolves added metadata by host (case-insensitive), overwrites and chains
  `AddOrUpdate`, falls back to the fallback provider then `null` for unknown hosts (a `null`/empty host resolves
  to `null` without throwing, short-circuiting before the fallback), and merges **per key**: a per-key override
  wins while sibling keys from the fallback are preserved, a per-key add keeps the fallback's keys, and two
  per-key calls accumulate distinct keys for the same host.
- The returned dictionary is a genuine read-only, case-insensitive view.
- Both `AddOrUpdate` overloads validate their arguments consistently with the MSAL and MISE twins: a null,
  empty, or whitespace `authorityHost` (and `key`, on the per-key overload) throws `ArgumentException`; a null
  values bag or value throws `ArgumentNullException`.


src/Microsoft.Identity.Abstractions/CloudMetadata/CloudMetadata.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <summary>
+        /// Gets the case-insensitive collection of cloud-specific key/value pairs held by this instance.
+        /// </summary>
+        public IReadOnlyDictionary<string, string> Values
+        {
+            get
+            {
+                return _values;
+            }
+        }
```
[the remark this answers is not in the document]
Avery-Dunn: _values  is now a  ReadOnlyDictionary<string,string>  and  Values  returns it, so callers can't downcast-and-mutate. Added  CloudMetadata_Values_IsReadOnly  to cover it.

src/Microsoft.Identity.Abstractions/CloudMetadata/InMemoryCloudMetadataProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+                throw new ArgumentNullException(nameof(authorityHost));
+            }
+
+            if (values is null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+#endif
+
+            ConcurrentDictionary<string, string> bag = GetOrAddBag(authorityHost);
```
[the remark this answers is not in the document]
Avery-Dunn: Both  AddOrUpdate  overloads now reject null/empty/whitespace  authorityHost  (and  key ) with  ArgumentException , and  GetByAuthorityHost(null/empty)  short-circuits to  null  before the fallback — consistent with the MSAL and MISE twins. Covered by the reworked  InMemoryProvider_InvalidArguments_Throw  tests.

src/Microsoft.Identity.Abstractions/CloudMetadata/AbstractionsCloudKeys.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// identical across every SDK that recognizes it, so a value written under a key by one layer can be
+    /// read under the same key by another layer without any translation table.
+    /// </para>
+    /// <para>
+    /// The list is <b>add-only</b>: new keys can be introduced without breaking existing callers, because
+    /// consumers look up only the keys they understand and ignore the rest. A key that must be retired
+    /// should first be marked <see cref="System.ObsoleteAttribute"/> for at least one release before removal.
+    /// </para>
+    /// </remarks>
+    public static class AbstractionsCloudKeys
```
bgavrilMS: Stange name. Why not just `CloudSettingKeyNames` ?

src/Microsoft.Identity.Abstractions/CloudMetadata/CloudMetadata.cs
```diff
@@ -0,0 +1,136 @@
...
+    ///     [AbstractionsCloudKeys.TokenExchangeAudience] = "api://AzureADTokenExchangeUSGov",
+    /// });
+    ///
+    /// if (metadata.TryGetValue(AbstractionsCloudKeys.TokenExchangeAudience, out string? audience))
+    /// {
+    ///     // use audience
+    /// }
+    /// </code>
+    /// </example>
+    public sealed class CloudMetadata
```
bgavrilMS: As mentioned in the MSAL PR - is this really needed? It's just a dictionary.

src/Microsoft.Identity.Abstractions/CloudMetadata/InMemoryCloudMetadataProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+    ///     {
+    ///         [AbstractionsCloudKeys.TokenExchangeAudience] = "api://AzureADTokenExchangeMyCloud",
+    ///     });
+    ///
+    /// services.AddSingleton&lt;ICloudMetadataProvider&gt;(provider);
+    /// </code>
+    /// </example>
+    public sealed class InMemoryCloudMetadataProvider : ICloudMetadataProvider
+    {
+        private readonly ConcurrentDictionary<string, ConcurrentDictionary<string, string>> _entries =
```
bgavrilMS: In MSAL you went with statics and ReadOnly dictionaries. Here you are going with ConcurrentDictionaries. Why?

src/Microsoft.Identity.Abstractions/CloudMetadata/InMemoryCloudMetadataProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            CloudMetadata? fallbackMetadata = _fallback?.GetByAuthorityHost(authorityHost);
+
+            if (!_entries.TryGetValue(authorityHost, out ConcurrentDictionary<string, string>? bag))
+            {
+                return fallbackMetadata;
+            }
+
+            return CloudMetadata.Merge(fallbackMetadata, new CloudMetadata(bag));
```
bgavrilMS: Why merge?

src/Microsoft.Identity.Abstractions/CloudMetadata/ICloudMetadataProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// implementation that knows about internal-only sovereign clouds, and Microsoft.Identity.Web resolves it
+    /// from the shared container to obtain the correct FIC token-exchange audience for those clouds.
+    /// </para>
+    /// <para>
+    /// Implementations should be thread-safe and return quickly, as they may be consulted on the token
+    /// acquisition hot path. Use <see cref="InMemoryCloudMetadataProvider"/> for a ready-made mutable
+    /// implementation.
+    /// </para>
+    /// </remarks>
+    public interface ICloudMetadataProvider
```
bgavrilMS, line 26: For now, we have only static / well-known providers. Think about what changes we will need to ensure we can have a provider that is async. Maybe we need some `InitializeAsync` method in this interface?

src/Microsoft.Identity.Abstractions/CloudMetadata/AbstractionsCloudKeys.cs
Avery-Dunn: Renamed to `CloudMetadataKeyNames` in the latest commit. Originally it had "Abstractions" in the name to avoid the risk of conflicting with other similar class names, and I kept the "Metadata" wording to be consistent with this feature's naming across each SDK.

src/Microsoft.Identity.Abstractions/CloudMetadata/CloudMetadata.cs
Avery-Dunn: Removed in the latest commit, as part of the simplification across the SDKs.

src/Microsoft.Identity.Abstractions/CloudMetadata/InMemoryCloudMetadataProvider.cs
Avery-Dunn: As of the latest commit it now follows the same static/readonly style.

src/Microsoft.Identity.Abstractions/CloudMetadata/InMemoryCloudMetadataProvider.cs
Avery-Dunn: This allows a caller to customize a single key in a particular cloud and inherit the rest of the metadata.

src/Microsoft.Identity.Abstractions/CloudMetadata/ICloudMetadataProvider.cs
Avery-Dunn: As far as I can tell, every provider today is an in-memory dictionary read, so making this async now doesn't seem necessary. If a real async metadata source appears we can add a separate async interface (or InitializeAsync on it) and have the resolver prefer it as an additive and non-breaking change.

src/Microsoft.Identity.Abstractions/CloudMetadata/ICloudMetadataProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Gets the cloud-specific metadata for the cloud that owns <paramref name="authorityHost"/>.
+        /// </summary>
+        /// <param name="authorityHost">The authority host (for example <c>login.microsoftonline.us</c>). This
+        /// is a bare host, not a full URL.</param>
+        /// <returns>A case-insensitive, read-only dictionary of cloud-specific values (keyed by
+        /// <see cref="CloudMetadataKeyNames"/>) for that cloud, or <c>null</c> if the host is unknown to this
+        /// provider (allowing a caller to fall back to another provider or a default). A <c>null</c> or empty
+        /// <paramref name="authorityHost"/> also resolves to <c>null</c> rather than throwing.</returns>
+        IReadOnlyDictionary<string, string>? GetByAuthorityHost(string authorityHost);
```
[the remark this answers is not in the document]
Avery-Dunn: Implemented in latest commit.

src/Microsoft.Identity.Abstractions/CloudMetadata/InMemoryCloudMetadataProvider.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+#endif
+
+            Dictionary<string, string> bag = GetOrAddBag(authorityHost);
+            foreach (KeyValuePair<string, string> pair in values)
+            {
+                bag[pair.Key] = pair.Value;
+            }
```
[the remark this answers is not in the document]
Avery-Dunn: Implemented in latest commit.

bgavrilMS (approved):

src/Microsoft.Identity.Abstractions/CloudMetadata/CloudMetadataKeyNames.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <para>
+    /// These string literals are the <b>shared vocabulary</b> used to exchange cloud-specific values
+    /// between cooperating SDKs and their callers. A key literal is intentionally
+    /// identical across every SDK that recognizes it, so a value written under a key by one layer can be
+    /// read under the same key by another layer without any translation table.
+    /// </para>
+    /// <para>
+    /// The list is <b>add-only</b>: new keys can be introduced without breaking existing callers, because
+    /// consumers look up only the keys they understand and ignore the rest. A key that must be retired
+    /// should first be marked <see cref="System.ObsoleteAttribute"/> for at least one release before removal.
```
trwalke: Even if the key is marked as obsolete, the removal of it will make it a breaking change. can they just be left forever?

Also, what if the customer upgrades past the version where the key was marked as obsolete, and right into the version where it is removed?

Perhaps I dont understand the word "removal" in this context?

Avery-Dunn: Yes, the original wording was misleading but I just updated it: when no longer needed keys can be marked as "obsolete" to discourage use, and we can (for example) set it to null to trigger fallback key-not-found behavior in our other SDKs without introducing a compile-time breakage to our customers.

trwalke (approved):

